PR #315, opened 2022-08-12
Release to fusion

einarmo:


cemreyavuz (approved):
Lots of black magic for me in the workflow definition - but LGTM

PR #319, opened 2022-09-26
[DOG-239] Added dockerhub tag for  bridge image

timurgen:


einarmo (approved):

einarmo:
You probably have to delete the tag and release from github if you want to re-release. The release in Fusion can stay, it will just be replaced.

PR #321, opened 2022-10-06
Redo ua trace

einarmo:
The SDK now allows writing traces directly to an ILogger, which greatly simplifies the code.

We would have probably done this slightly differently if this was supported from the beginning, but it is not worth breaking backwards compatibility of the config file to clean it up.


timurgen (approved):
lgtm!

PR #323, opened 2022-10-19
Refactor connection logic

einarmo:
The idea is to lower restart times and also clean up the code a bit. Instead of restarting from zero when the extractor fails to connect, we restart inside a new SessionManager class.

This PR also improves the ForceRestart option to not quit the extractor entirely, which greatly improves restart time. Some users have been hesitant to use the ForceRestart option because it is so slow, this should help them considerably.

This is also the groundwork for supporting redundant servers. For those it is nice to have some abstraction around the session, since it may represent one of several servers.

### DIFF
--- a/ConfigurationTool/Checks/Attributes.cs
+++ b/ConfigurationTool/Checks/Attributes.cs
@@ -47,7 +47,7 @@ namespace Cognite.OpcUa.Config
 
             if (Session == null || !Session.Connected)
             {
-                await Run(token);
+                await Run(token, 0);
                 await LimitConfigValues(token);
             }
 
--- a/ConfigurationTool/Checks/Browse.cs
+++ b/ConfigurationTool/Checks/Browse.cs
@@ -133,7 +133,7 @@ namespace Cognite.OpcUa.Config
         {
             if (Session == null || !Session.Connected)
             {
-                await Run(token);
+                await Run(token, 0);
                 await LimitConfigValues(token);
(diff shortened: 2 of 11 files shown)


ConfigurationTool/Checks/Attributes.cs
```diff
@@ -47,7 +47,7 @@ public async Task GetAttributeChunkSizes(CancellationToken token)
 
             if (Session == null || !Session.Connected)
             {
-                await Run(token);
+                await Run(token, 0);
```
timurgen, line 50: What does 0 mean here? If 0 is timeout isn't it too short for session establishing purpose?

timurgen: Also shouldn't it be a configurable option to adjust to different needs?

einarmo: Hm. There is no real advantage to timing out reconnects for the extractor, since it restarts on crash anyway. I added this because we rely so heavily on the Run method failing in tests. I.e. letting it be 0 for tests, and leaving it at -1 for normal operation is fine.

The logic behind 0 is that the timeout is checked _after_ a failed reconnect attempt. So it tries to connect, and if it fails it checks if `timeout` has elapsed, and rethrows the exception if it has. So 0 just means always rethrow the exception.

timurgen (approved):
lgtm!

PR #325, opened 2022-10-24
[DOG-363] Redundancy support

einarmo:
Support redundant OPC-UA servers, using non-transparent redundancy.

This adds an `alt-endpoint-urls` property containing a list of URLs. The one with the lowest read `ServiceLevel` is selected.


toondaey (commented):
🚀

toondaey (approved):

PR #330, opened 2022-11-04
Better handling of certificate files in docker

einarmo:
I'd like to find a better way to handle these kinds of special configs in docker, and in general overrides for the default xml config.


timurgen (approved):
lgtm!

PR #332, opened 2022-11-15
Retries

einarmo:
Basic system for retries on OPC-UA service calls. This does not handle every case, and can have somewhat unpredictable results if the server is poorly behaved, but it seems to work well most of the time.


toondaey (approved):
lgtm

PR #334, opened 2022-11-25
Use _PUBLIC_REPOS variants of artifactory tokens

einarmo:
When security was making these available for every repo, it seems like they made them unavailable for public repos.


timurgen (approved):
lgtm!

PR #335, opened 2023-01-06
Fix retries

einarmo:
The issue was that when silenced exceptions were given to HandleServiceResult, they were treated as unknown exceptions. We would rather just return them back.

This also contains a bit of server code to make testing this easier, though it's not something I would like to include in tests.

It also adds `BadNoCommunication` as a default status code.

### DIFF
--- a/Extractor/UAClient.cs
+++ b/Extractor/UAClient.cs
@@ -283,6 +283,12 @@ namespace Cognite.OpcUa
                     return Config.Source.Retries.FinalRetryStatusCodes.Contains(silentExc.InnerServiceException.StatusCode);
                 }
             }
+            else if (ex is AggregateException aex)
+            {
+                // Only retry aggregate exceptions if all inner exceptions should be retried...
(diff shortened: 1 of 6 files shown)


Extractor/UAClient.cs
```diff
@@ -283,6 +283,12 @@ private bool ShouldRetryException(Exception ex)
                     return Config.Source.Retries.FinalRetryStatusCodes.Contains(silentExc.InnerServiceException.StatusCode);
                 }
             }
+            else if (ex is AggregateException aex)
+            {
+                // Only retry aggregate exceptions if all inner exceptions should be retried...
```
mathialo, line 288: Only retry if all inner should be retried? If any should be retried, should we not retry that one?

einarmo: An aggregate exception technically means that multiple things have gone wrong... I guess we _could_ retry, I'm not 100% sure.

PR #336, opened 2023-01-09
Pass along whether a node source result can be used for deletes

einarmo:
Also add a few config options we'll need.

This is the first small step towards deletes. We need to pass along this information so that partial browses (triggered by audit events or the like) do not make us delete everything in CDF. (though we probably won't implement hard deletes at all)


toondaey (approved):
👍🏾

PR #339, opened 2023-01-17
Trigger rebrowse based on updates to the server namespaces node(s)

toondaey:


Extractor/Config.cs
```diff
@@ -642,6 +642,19 @@ public class SubscriptionConfig
...
+        /// </summary>
+        public ServerNamespacesToRebrowseConfig? ServerNamespacesToRebrowse { get; set; }
+    }
+    public class ServerNamespacesToRebrowseConfig
+    {
+        /// <summary>
+        /// Desired namespace nodes to subscribe to.
+        /// </summary>
+        /// <typeparam name="ProtoNodeId"></typeparam>
+        public IEnumerable<ProtoNodeId> NamespaceNodes { get; set; } = new List<ProtoNodeId>();
```
einarmo, line 657: This is kinda silly. Why do we need to specify the NodeId of each namespace?

First of all, does it not make sense to just subscribe to all of them? For users like the one we met who had hundreds of namespaces, this may be quite clunky.

Secondly, namespaces have unique names, it is cleaner to just use those, rather than needing to note down the actual NodeId of each.

Extractor/ServerSubscriptionManager.cs
```diff
@@ -0,0 +1,105 @@
...
+            {
+                var nc = (NodeClass)results[id * attributes.Count() + 3].Value;
+                if (nc != NodeClass.Variable) continue;
+                var rawDt = results[id * attributes.Count() + 2].GetValue(NodeId.Null);
+
+                nodes[nodeIds[id]] = new ServerItemSubscriptionState(_uaClient, nodeIds[id]);
+            }
+
+            await _uaClient.AddSubscriptions(nodes.Values, "NodeMetrics",
+               async (MonitoredItem item, MonitoredItemNotificationEventArgs _) => await extractor.Rebrowse(),
```
einarmo, line 84: This is extremely dangerous for multiple reasons.

First of all, this blocks the entire event queue, so no new data arrives while Rebrowse is running, which remember, can take a long time.

Secondly, if you receive 100 notifications at the same time here, which isn't actually that unlikely, you will rebrowse the OPC-UA server _100_ times, which would be bad.

Instead, do as we discussed and trigger rebrowse on the looper. This does not block, and also does not run more than once in parallel, if new notifications arrive while rebrowse is running, it will simply browse once more when it is done.

Extractor/UAClient.cs
```diff
@@ -117,6 +123,12 @@ public async Task Run(CancellationToken token, int timeout = -1)
             await StartSession(timeout);
             await StartNodeMetrics();
         }
+
+        public async Task StartCustomServerSubscriptions(UAExtractor extractor)
```
einarmo, line 127: Does this need to be public at all? When are the node metrics subscriptions created?

Extractor/ServerSubscriptionManager.cs
einarmo: It also does not take into consideration that you will receive notifications for the initial value of the node.

Extractor/UAClient.cs
```diff
@@ -104,6 +106,10 @@ public UAClient(IServiceProvider provider, FullConfig config)
             {
                 metricsManager = new NodeMetricsManager(this, config.Source, config.Metrics.Nodes);
             }
+            serverSubscription = new ServerSubscriptionManager(
```
einarmo, line 109: We probably do not need to create this if the system for subscribing to changes is not enabled.

Extractor/ServerSubscriptionManager.cs
```diff
@@ -0,0 +1,105 @@
...
+
+using Cognite.OpcUa.History;
+using Microsoft.Extensions.Logging;
+
+using Opc.Ua;
+using Opc.Ua.Client;
+
+namespace Cognite.OpcUa
+{
+    public class ServerSubscriptionManager
```
einarmo, line 14: The name isn't great.

Extractor/ServerSubscriptionManager.cs
```diff
@@ -0,0 +1,105 @@
...
+                            refDef.DisplayName.ToString(), nodeId
+                        );
+                    }
+                },
+                token,
+                maxDepth: -1,
+                doFilter: false
+            );
+
+            var readValueIds = new ReadValueIdCollection(
```
einarmo, line 66: Do we actually need to read this?

Extractor/ServerSubscriptionManager.cs
```diff
@@ -0,0 +1,105 @@
...
+
+                nodes[nodeIds[id]] = new ServerItemSubscriptionState(_uaClient, nodeIds[id]);
+            }
+
+            await _uaClient.AddSubscriptions(nodes.Values, "NodeMetrics",
+               async (MonitoredItem item, MonitoredItemNotificationEventArgs _) => await extractor.Rebrowse(),
+                state => new MonitoredItem
+                {
+                    StartNodeId = state.SourceId,
+                    SamplingInterval = 100,
```
einarmo, line 88: We probably don't need a sampling interval this aggressive here. This means sample 10 times a second.

einarmo (commented):
See comments, it's a start, but I disagree with some design decisions, and I'm not sure it will work as intended.

Extractor/Config.cs
toondaey: I understand that users may have a lot of namespaces but giving them the option to provide a list of namespaces to subscribe to will future-proof this feature.
Personally, I think it would be better for users to provide the node id each namespace that way it's clarity as to what the user will be subscribing to.

Extractor/ServerSubscriptionManager.cs
toondaey: I forgot to change this after completing the code.

Extractor/ServerSubscriptionManager.cs
toondaey: I had to do this because I noticed that the `AddSubscriptions` Method takes a list of `UAHistoryExtractionState`.

Extractor/ServerSubscriptionManager.cs
toondaey: Will work on coming up with a more descriptive name.

Extractor/ServerSubscriptionManager.cs
toondaey: I'll work on moving this to the looper class.

Extractor/UAClient.cs
einarmo: I'm not a huge fan of that logic. I think we should try to avoid this kind of circular references as much as possible, where

 - UAClient owns a class
 - That class owns a reference to UAExtractor
 - UAExtractor owns UAClient

Is there a reason why UAClient should own the subscription handler and not UAExtractor? Since this affects extractor behavior I think it belongs on the extractor more than the client.

Extractor/Config.cs
einarmo: To me it kind of feels like it does the opposite of future-proofing it, i.e. we make assumptions about the usage that have not yet been proven? Any reason why we don't just have a true/false now, then optionally add a filter in the future?

Extractor/Config.cs
```diff
@@ -642,6 +642,23 @@ public class SubscriptionConfig
         /// Log bad subscription datapoints
         /// </summary>
         public bool LogBadValues { get; set; } = true;
+
+        /// <summary>
+        /// Server namespace nodes the should be subscribed to a rebrowse upon changes to their values.
+        /// </summary>
+        public ServerNamespacesToRebrowseConfig? ServerNamespacesToRebrowse { get; set; }
```
einarmo, line 649: Maybe change the name in config? Since `Namespace` is already an OPC-UA concept, we don't really need to specify `ServerNamespace`.

Maybe we future proof it, and say something like `RebrowseTriggers`, or something?

Extractor/Config.cs
```diff
@@ -642,6 +642,23 @@ public class SubscriptionConfig
...
+    }
+    public class ServerNamespacesToRebrowseConfig
+    {
+        /// <summary>
+        /// Indicate if the extractor should turn on subscription to namespaces.
+        /// </summary>
+        public bool Subscribe { get; set; } = false;
```
einarmo, line 656: That would also mean changing this to be slightly more specific, maybe `namespace-publication-date` or something.

Extractor/Config.cs
```diff
@@ -642,6 +642,23 @@ public class SubscriptionConfig
...
+        /// <summary>
+        /// Desired namespace nodes to subscribe to.
```
einarmo, line 658: And then have this be optional. Be aware that setting it equal to something here does not prevent a user from setting it to null in config, so you can't trust the compiler now telling you that it is non-null.

Extractor/ServerSubscriptionManager.cs
```diff
@@ -0,0 +1,106 @@
...
+
+                nodes[nodeIds[id]] = new ServerItemSubscriptionState(_uaClient, nodeIds[id]);
+            }
+
+            await _uaClient.AddSubscriptions(nodes.Values, "NodeMetrics",
+               (MonitoredItem item, MonitoredItemNotificationEventArgs _) => extractor.Looper.QueueRebrowse(),
+                state => new MonitoredItem
+                {
+                    StartNodeId = state.SourceId,
+                    SamplingInterval = 10,
```
einarmo, line 89: This is even lower, isn't it? I think this is in milliseconds, so now it's 100 times a second, might be a bit much.

Extractor/ServerSubscriptionManager.cs
einarmo: I'm not actually 100% sure it needs to, we should maybe investigate that a bit later.

Extractor/ServerSubscriptionManager.cs
```diff
@@ -0,0 +1,106 @@
...
+            Attributes.DisplayName,
+            Attributes.DataType,
+            Attributes.NodeClass,
+            Attributes.Description
+        };
+
+        public ServerSubscriptionManager(
+            ILogger<ServerSubscriptionManager> logger,
+            UAClient uaClient,
+            ServerNamespacesToRebrowseConfig? config
```
einarmo, line 32: I think you should be able to assume the config to be non-null here.

einarmo (commented):
Getting there!

Extractor/ServerSubscriptionManager.cs
einarmo: It's better now, but we should look into preventing rebrowse being triggered from the initial value. Perhaps you can check if the data point received has a source timestamp/server timestamp greater than the start of the extractor? No point browsing for an update that we have already covered.

Extractor/Config.cs
toondaey: This would indicate that we only allow rebrowse triggers on specific nodes?

einarmo: Yeah, I figured we can add systems to subscribe to more nodes later. So have

`namespace-publication-date` subscribe to all namespace publication date nodes, unless filtered by `namespaces`, then we could add, in the future, a list of additional NodeIds that the user can subscribe to.

toondaey: Got it.

toondaey: If each node id is unique and we know the nodes we're targeting (e.g. with `namespace-publication-date`), I don't think we still need the `namespaces` filter. What do you think?

Extractor/Config.cs
```diff
@@ -642,7 +640,43 @@ public class SubscriptionConfig
...
+                if (value && !exists) {
+                    _toBeSubscribed.Add("NamespacePublicationDate");
+                } else if (!value && exists) {
+                    _toBeSubscribed.Remove("NamespacePublicationDate");
+                }
+            }
+        }
+
+        // TODO: rename this method appropriately.
+        public List<string> GetValues => _toBeSubscribed;
```
einarmo, line 677: Two minor things here. This project does not use `_` prefixes for private members, (I know it's convention, but consistency trumps). Also it is common to use a similar name for the public property as the private member, to make the relationship clear, i.e. this should probably be `ToBeSubscribed`.

Extractor/ServerSubscriptionManager.cs
```diff
@@ -0,0 +1,153 @@
...
+            var processedNamespaces = (
+                shouldFilterNamespaces
+                    ? filteredNamespaces.Intersect(availableNamespaces)
+                    : availableNamespaces
+            ).ToList();
+
+            if (shouldFilterNamespaces && processedNamespaces.Count() < filteredNamespacesCount)
+            {
+                _logger.LogInformation(
+                    "Some namespaces could not be processed as they do not exist on the server: {namespaces}",
```
einarmo, line 90: Maybe use something other than "processesed"? Remember that users generally don't have the context of where the log message comes from.

Extractor/Config.cs
```diff
@@ -642,7 +640,43 @@ public class SubscriptionConfig
         /// Log bad subscription datapoints
         /// </summary>
         public bool LogBadValues { get; set; } = true;
+
+        /// <summary>
+        /// Server namespace nodes the should be subscribed to a rebrowse upon changes to their values.
+        /// </summary>
+        public RebrowseTriggersConfig? RebrowseTriggers { get; set; }
```
einarmo, line 647: Thinking about it, I'm not so sure this belongs in `Subscriptions`, yes it does create subscriptions, but this section is generally more about modifying the data subscriptions we make. Things related to rebrowse are generally under `Extraction`.

Extractor/ServerSubscriptionManager.cs
```diff
@@ -0,0 +1,153 @@
...
+        private async Task CreateSubscriptions(List<ServerItemSubscriptionState> nodes, CancellationToken token)
+        {
+            var sub = await _uaClient.AddSubscriptions(
+                nodes, "TriggerRebrowse",
+                async (MonitoredItem item, MonitoredItemNotificationEventArgs _) =>
+                {
+                    var readValueId = new [] { 
+                        new ReadValueId { NodeId = item.ResolvedNodeId, AttributeId = Attributes.Value }
+                    };
+                    var read = await _uaClient.ReadAttributes(new ReadValueIdCollection(readValueId), 1, token);
```
einarmo, line 122: This you don't need to do. See how this is done in `Streamer`. You can use `DequeueValues`, the values are passed along with the notification.

einarmo (commented):
Getting closer, I think, just a few small things here, then of course you need error handling, tests, and to fix the compiler warnings.

schema/subscription_config.schema.json
```diff
@@ -46,6 +46,25 @@
             "type": "boolean",
             "description": "Log bad subscription datapoints",
             "default": true
+        },
+        "server-namespaces-to-rebrowse-config": {
```
einarmo, line 50: Looks like you have to update this, to match the moved config.

Extractor/UAExtractor.cs
```diff
@@ -287,7 +299,17 @@ private async Task RunExtractorInternal(int startTimeout = -1)
             }
 
             Started = true;
-            startTime.Set(new DateTimeOffset(DateTime.UtcNow).ToUnixTimeMilliseconds());
+
+            if (
+                serverSubscription is not null
+                && Config.Extraction.RebrowseTriggers is not null
```
einarmo, line 305: One thing we do quite a lot is to just check if the manager type is null, because we only create it if we are going to need it. Since we check `config.Extraction.RebrowseTriggers` when creating the manager, we don't need to check it here.

Extractor/UAExtractor.cs
```diff
@@ -669,6 +691,13 @@ private async Task EnsureSubscriptions()
                 await uaClient.SubscribeToNodes(subscribeStates, Streamer.DataSubscriptionHandler, Source.Token);
             }
 
+            if (
+                serverSubscription is not null
+                && Config.Extraction.RebrowseTriggers is not null
```
einarmo, line 696: or here

Extractor/UAExtractor.cs
toondaey: I also checked the `serverSubscription` because the compiler throws this warning `Dereference of a possibly null reference. [Extractor]` without it.

Extractor/UAExtractor.cs
toondaey: Same here

Extractor/UAExtractor.cs
einarmo: I mean, check only serverSubscription. For the nullability warning, you can ignore it here. Use the `!` operator, if you need to.

Extractor/Config.cs
```diff
@@ -643,6 +645,39 @@ public class SubscriptionConfig
...
+
+                if (value && !exists) {
+                    ToBeSubscribed.Add("NamespacePublicationDate");
+                } else if (!value && exists) {
+                    ToBeSubscribed.Remove("NamespacePublicationDate");
+                }
+            }
+        }
+
+        // TODO: rename this method appropriately.
```
einarmo, line 677: Resolve this TODO?

Extractor/RebrowseTriggerManager.cs
```diff
@@ -0,0 +1,151 @@
...
+            await _uaClient.Browser.BrowseDirectory(
+                new[] { serverNamespaces },
+                (refDef, parent) =>
+                {
+                    var nodeId = (NodeId)refDef.NodeId;
+
+                    if (parent == serverNamespaces && !grouping.ContainsKey(nodeId))
+                    {
+                        grouping.Add(nodeId, new List<ReferenceDescription>());
+                        namespaceNameToId.Add(refDef.DisplayName.ToString(), nodeId.ToString());
```
einarmo, line 59: I'd rather we don't use NodeId.ToString(): Is there are reason why you need it here? NodeId implements equality properly, so you can just use the `NodeId` class directly.

Extractor/RebrowseTriggerManager.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        private async Task CreateSubscriptions(List<ServerItemSubscriptionState> nodes, CancellationToken token)
+        {
+            var sub = await _uaClient.AddSubscriptions(
+                nodes, SubscriptionName,
+                (MonitoredItem item, MonitoredItemNotificationEventArgs _) =>
+                {
+                    var values = item.DequeueValues();
+                    var value = values.Count > 0 
+                        ? values[0].GetValue<System.DateTime>(UAExtractor.StartTime)
```
einarmo, line 122: Should we instead of using StartTime keep the time we last triggered a rebrowse, and default to the startup time? That might be a bit more consistent, in case we get multiple simultaneous notifications.

Extractor/UAExtractor.cs
```diff
@@ -95,6 +95,10 @@ public class UAExtractor : BaseExtractor<FullConfig>, IUAClientAccess
 
         private readonly ILogger<UAExtractor> log;
 
+        private readonly RebrowseTriggerManager? serverSubscription;
```
einarmo, line 98: Rename the manager here as well, just `rebrowseTriggerManager` or something is fine.

ExtractorLauncher/ExtractorStarter.cs
```diff
@@ -79,6 +79,14 @@ public static class ExtractorStarter
                 var parsed = CogniteTime.ParseTimestampString(config.History.EndTime);
                 if (parsed == null) return $"Invalid history end time: {config.History.EndTime}";
             }
+
+            if (
+                config.Extraction.RebrowseTriggers is not null 
+                && config.Extraction.RebrowseTriggers.Targets == null
```
einarmo, line 85: This is impossible, isn't it?

Extractor/RebrowseTriggerManager.cs
```diff
@@ -0,0 +1,151 @@
...
+                    }
+                    else if (
+                        grouping.ContainsKey(parent)
+                        // Ensures that the type of node being added is a variable node class
+                        && refDef.NodeClass == NodeClass.Variable
+                        // Filters targets nodes
+                        && targetNodes.Contains(refDef.DisplayName.ToString())
+                    )
+                    {
+                        grouping.GetValueOrDefault(parent).Add(refDef);
```
einarmo, line 69: This can be done with TryGetValue?

schema/extraction_config.schema.json
```diff
@@ -319,6 +319,25 @@
                     "description": "Another filter instance which is applied to the parent node"
                 }
             }
+        },
+        "rebrowse-triggers": {
+            "type": "object",
+            "unevaluatedProperties": false,
+            "description": "Subscribes server namespace nodes to rebrowse which can be filtered by namespaces.",
```
einarmo, line 326: This is a bit clunky, maybe "Configuration for triggering rebrowse based on changes to specific nodes".

schema/extraction_config.schema.json
```diff
@@ -319,6 +319,25 @@
...
+            "properties": {
+                "namespace-publication-date": {
+                    "type": "boolean",
+                    "description": "Indicate if the extractor should turn on subscription to namespaces."
```
einarmo, line 330: "Indicate if the extractor should subscribe to namespacePublicationDate on the Server node."

schema/extraction_config.schema.json
```diff
@@ -319,6 +319,25 @@
...
+                },
+                "namespaces": {
+                    "type": "array",
+                    "description": "A list of server namespaces by which we should filter the nodes.",
```
einarmo, line 334: "A list of namespaces used to filter which namespacePublicationDates the extractor will subscribe to"

schema/extraction_config.schema.json
```diff
@@ -319,6 +319,25 @@
...
+                    "items": {
+                        "type": "string",
+                        "description": "The display names of the namespaces."
```
einarmo, line 337: Singular here,

"Display name of a namespace"

einarmo (commented):
The configuration options should also be added to the example config, `config/config.example.yml`, you can just reuse the descriptions from the schema.

config/config.example.yml
```diff
@@ -765,6 +765,15 @@ extraction:
             # historizing:
             # Another instance of NodeFilter which is applied to the parent node.
             # parent:
+    # Configure extractor to trigger a rebrowse when there are changes to specific namespace metadata nodes.
+    # Also supports filtering by namespace uris.
+    rebrowse-triggers:
+        # A dictionary of nodes to which we would like to listen for changes.
+        targets:
+            namespace-publication-date: true
```
einarmo, line 773: It's false by default, so set it to false here.

einarmo (approved):
I had one final comment, but apart from that it looks good. Great work.

PR #341, opened 2023-01-18
Improve handling of missing namespace again

einarmo:
This should make it work, and communicate clearly to the user that there is a problem.

### DIFF
--- a/Extractor/UAClient.cs
+++ b/Extractor/UAClient.cs
@@ -1477,7 +1477,7 @@ namespace Cognite.OpcUa
                 }
                 else
                 {
-                    log.LogWarning("NodeID received with null NamespaceUri, and index not in NamespaceTable. This should not be possible! {Id}, {Index}",
+                    log.LogWarning("NodeID received with null NamespaceUri, and index not in NamespaceTable. This is likely a bug in the server. {Id}, {Index}",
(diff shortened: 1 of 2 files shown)


Extractor/UAClient.cs
```diff
@@ -1463,9 +1463,24 @@ public static double ConvertToDouble(object datavalue)
...
+                        newPrefix = prefixNode;
+                    }
+                    else
+                    {
+                        newPrefix = namespaceUri + ":";
+                    }
+                }
+                else
+                {
+                    log.LogWarning("NodeID received with null NamespaceUri, and index not in NamespaceTable. This should not be possible! {Id}, {Index}",
```
toondaey, line 1480: I think this should be more descriptive .
```suggestion
                    log.LogWarning("NodeID received with null NamespaceUri, and index not in NamespaceTable. Unprocessable namepace! {Id}, {Index}",
```

toondaey (approved):
looks good

Extractor/UAClient.cs
einarmo: That doesn't seem more descriptive as much as the same with more technical language, but I can agree that the wording isn't very clear. I wanted to avoid pointing fingers too hard. Perhaps `This is likely a bug in the server`?

PR #342, opened 2023-01-19
Store known nodes

einarmo:
System to store known nodes and identify which has been deleted. This does not yet do anything in the extractor, it is just utility code that we will use to mark deletes later.

### DIFF
--- a/Extractor/Deletes.cs
+++ b/Extractor/Deletes.cs
@@ -73,7 +73,7 @@ namespace Cognite.OpcUa
             if (deleted.Any())
             {
                 logger.LogInformation("Found {Del} stored nodes in {Tab} that no longer exist and will be marked as deleted", deleted.Count, tableName);
-                await stateStore.DeleteExtractionState(deleted.Cast<IExtractionState>(), tableName, token);
+                await stateStore.DeleteExtractionState(deleted.Select(d => new NodeExistsState(d.Id)), tableName, token);
             }
 
             var newStates = states.Values.Where(s => !oldStates.ContainsKey(s.Id)).ToList();
@@ -81,9 +81,8 @@ namespace Cognite.OpcUa
             if (newStates.Any())
             {
                 logger.LogInformation("Found {New} new nodes in {Tab}, adding to state store...", newStates.Count, tableName);
-                await stateStore.StoreExtractionState(newStates, tableName, token);
+                await stateStore.StoreExtractionState(newStates, tableName, s => new BaseStorableState { Id = s.Id }, token);
             }
-
 
             return deleted.Select(d => d.Id);
         }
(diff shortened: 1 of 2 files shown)


Extractor/Deletes.cs
```diff
@@ -0,0 +1,108 @@
+﻿using Cognite.Extractor.StateStorage;
+using Cognite.OpcUa.NodeSources;
+using Microsoft.Extensions.Logging;
+using Opc.Ua;
```
toondaey, line 4: Looks like this is unused

Extractor/Deletes.cs
```diff
@@ -0,0 +1,108 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
```
toondaey, line 8: Same here.

Test/Unit/DeleteTest.cs
```diff
@@ -0,0 +1,245 @@
...
+using Opc.Ua;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Test.Utils;
+using Xunit;
+using Xunit.Abstractions;
+using Xunit.Sdk;
```
toondaey, line 17: Looks like this is unused too.

Extractor/Deletes.cs
```diff
@@ -0,0 +1,108 @@
...
+            try
+            {
+                oldStates = (await stateStore.GetAllExtractionStates<BaseStorableState>(tableName, token)).ToDictionary(s => s.Id);
+            }
+            catch (Exception ex)
+            {
+                logger.LogWarning(ex, "Failed to get states from state store, assuming it is empty");
+                oldStates = new Dictionary<string, BaseStorableState>();
+            }
+            var deleted = oldStates.Where(s => !states.ContainsKey(s.Key)).Select(kvp => kvp.Value).ToList();
```
toondaey, line 71: Rather than creating a dangling variable, why don't we extract this into its own method (maybe `GetDeletedItems`). Makes it easier to read too.

toondaey (changes requested):

Extractor/Deletes.cs
einarmo: That seems a little unpleasant? We have to pass both `states` and `oldStates` to that method, and it will be a one-liner. I don't really see how that is clearer?

einarmo: I gave it a shot, but it takes up twice as many lines of code, does not reduce duplication, and IMO does not make the code more readable.

toondaey: The method can return both the deleted and the new items so it represents a sort of separation of states.

einarmo: I'm not sure I understand what you are asking for?

PR #344, opened 2023-01-25
Pass delete request to pushers

einarmo:
This ended up being somewhat more complicated than expected. The PushNodes methods in UAExtractor were not really that well suited to get a list of nodes to delete as well, and required a small rewrite. The system for caching nodes that failed during the initial push also required a little work. The result is (IMO) a lot cleaner, and makes a lot more sense.

Beyond that we now have an ExecuteDeletes method on IPusher, which can be implemented for CDFPusher to handle deletes.

### DIFF
--- a/Extractor/Pushers/IPusher.cs
+++ b/Extractor/Pushers/IPusher.cs
@@ -16,6 +16,7 @@ along with this program; if not, write to the Free Software
 Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA. */
 
 using Cognite.OpcUa.History;
+using Cognite.OpcUa.NodeSources;
 using Cognite.OpcUa.Types;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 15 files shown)


Extractor/Pushers/IPusher.cs
```diff
@@ -116,6 +127,17 @@ public interface IPusher : IDisposable
             return Task.FromResult((bool?)true);
         }
 
+        /// <summary>
+        /// Mark the given nodes as deleted in destinations.
+        /// </summary>
+        /// <param name="deletes"></param>
+        /// <param name="token"></param>
+        /// <returns></returns>
+        Task<bool> ExecuteDeletes(DeletedNodes deletes, CancellationToken token)
```
toondaey, line 136: Looks like neither of the 2 parameters here were used.

toondaey (approved):
looks fine but noticed some unused params.

Extractor/Pushers/IPusher.cs
einarmo: This is the default implementation of an interface.

PR #345, opened 2023-02-01
rebrowse trigger release manifest

toondaey:


manifest.yml
```diff
@@ -60,6 +60,11 @@ schema:
    - "https://raw.githubusercontent.com/"
 
 versions:
+  "2.11.0":
+    description: Improve handling of unknown namespace index
```
einarmo, line 64: Don't think this is correct.

manifest.yml
```diff
@@ -60,6 +60,11 @@ schema:
...
+    changelog:
+      added:
+        - "Adds support for rebrowse triggers"
```
einarmo, line 67: This is exposed to users, so it should be more descriptive. Perhaps "Adds support for triggering rebrowse based on updates to namespacePublicationDate"

einarmo (changes requested):

einarmo (approved):

PR #346, opened 2023-02-01
Delete in cdf

einarmo:
Adds support for deletes to the CDF Pusher. Relationships cannot be soft deleted in any practical way, so there is a separate option to hard-delete them.


toondaey (approved):
lgtm

PR #359, opened 2023-02-10
Update to .NET 7

einarmo:


toondaey (approved):
I think this is a function of how much we trust our tests and as such if the tests seem to be passing and not janky, I think it's good to go.

PR #360, opened 2023-02-15
Misc improvements to the service manager

einarmo:
It was not entirely clear where it was trying to find the executable, this makes it clearer, and also adds some alternative options.


toondaey (approved):
lgtm!

PR #363, opened 2023-02-24
Handle skipping a variable that is parent of another node

einarmo:
We skip variables for a variety of reasons. If MapVariableChildren is enabled, variables with children are turned into both an object and a variable, i.e.

```
Parent
  MyVariable
```

is turned into

```
Parent
  MyVariable (object)
    MyVariable
    Children of MyVariable
```

If we skipped MyVariable in AllowTsMap, we would skip the entire node. This isn't ideal, as we would like the result to be

```
Parent
  MyVariable (object)
    Children of MyVariable
```

So that children are mapped.

This solves that. The fix is very simple, just move the call to AllowTsMap into AddVariableToLists, only adding the variable to _variable_ lists if AllowTSMap returns true, but always adding it to _object_ lists.

### DIFF
--- a/Extractor/Types/UAVariable.cs
+++ b/Extractor/Types/UAVariable.cs
@@ -224,10 +224,10 @@ namespace Cognite.OpcUa.Types
 
         public struct VariableGroups
         {
-            public bool SourceObject;
-            public bool SourceVariable;
-            public bool DestinationObject;
-            public bool DestinationVariable;
+            public bool IsSourceObject;
+            public bool IsSourceVariable;
+            public bool IsDestinationObject;
+            public bool IsDestinationVariable;
         }
 
         public VariableGroups GetVariableGroups(DataTypeManager dataTypeManager)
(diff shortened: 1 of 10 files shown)


Extractor/Types/UAVariable.cs
```diff
@@ -212,7 +212,41 @@ public IEnumerable<UAVariable> CreateTimeseries()
...
+                SourceObject = NodeClass != NodeClass.Variable,
+                // Source variable if we wish to subscribe to it
+                SourceVariable = allowTsMap && NodeClass == NodeClass.Variable,
+                // Destination object if it's an object directly (through isObject)
+                // it's a mapped array, or it's not a variable.
+                DestinationObject = IsArray && Index == -1 && allowTsMap
+                    || isObject
+                    || NodeClass != NodeClass.Variable,
+                // Destination variable if allowTsMap is true and it's a variable.
+                DestinationVariable = allowTsMap && NodeClass == NodeClass.Variable,
```
toondaey, line 248: Can we make the variables more descriptive?
```suggestion
                IsSourceObject = NodeClass != NodeClass.Variable,
                // Source variable if we wish to subscribe to it
                IsSourceVariable = allowTsMap && NodeClass == NodeClass.Variable,
                // Destination object if it's an object directly (through isObject)
                // it's a mapped array, or it's not a variable.
                IsDestinationObject = IsArray && Index == -1 && allowTsMap
                    || isObject
                    || NodeClass != NodeClass.Variable,
                // Destination variable if allowTsMap is true and it's a variable.
                IsDestinationVariable = allowTsMap && NodeClass == NodeClass.Variable,
```

toondaey (changes requested):
👍🏾. Just some changes that should probably be considered...

Extractor/Types/UAVariable.cs
toondaey: I just felt that something like `VariableGroups.IsSourceObject` makes it more expressive.

toondaey (approved):
👍🏾

PR #364, opened 2023-02-27
Workaround for servers behind NAT

einarmo:
NAT (Network Address Translation) lets a server appear with a different discovery URL to the client than what it sees itself. Since the server likely doesn't know about this, the endpoints it exposes through discovery may have wrong host, port, etc.

This lets users override the returned endpoint URL.


toondaey (approved):
👍🏾
No manifest update though.

PR #367, opened 2023-03-02
Split config

einarmo:
The Config.cs file had become very large, this splits it into separate files roughly by top-level config object.


toondaey (approved):
lgtm!

PR #369, opened 2023-03-08
Monitor service level

einarmo:
This ended up being a fairly complex feature.

The main goal of this is to monitor the server ServiceLevel, trigger callbacks once it drops below a certain level, and not update extractor state if the server is in a bad state.

This effectively required better support for asynchronous callbacks. Using events had become somewhat clunky and hard to work with, it also caused us to do a lot of `var _ = Task.Run(...)`, which isn't ideal as it makes it so that exceptions are completely lost.


toondaey (approved):
👍🏾

PR #370, opened 2023-03-09
Add system for varying subscription configs per node

einarmo:
Works by letting users provide a collection of config, filter pairs, and having the first match be applied for each node.

### DIFF
--- a/ExtractorLauncher/ExtractorStarter.cs
+++ b/ExtractorLauncher/ExtractorStarter.cs
@@ -79,6 +79,16 @@ namespace Cognite.OpcUa
                 var parsed = CogniteTime.ParseTimestampString(config.History.EndTime);
                 if (parsed == null) return $"Invalid history end time: {config.History.EndTime}";
             }
+            if (config.Source.SamplingInterval != null)
+            {
+                log.LogWarning("source.sampling-interval is deprecated. Use subscriptions.sampling-interval instead.");
+                config.Subscriptions.SamplingInterval = config.Source.SamplingInterval.Value;
+            }
+            if (config.Source.QueueLength != null)
+            {
+                log.LogWarning("source.queue-length is deprecated. Use subscriptions.queue-length instead.");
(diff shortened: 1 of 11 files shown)


ExtractorLauncher/ExtractorStarter.cs
```diff
@@ -79,6 +79,16 @@ public static class ExtractorStarter
                 var parsed = CogniteTime.ParseTimestampString(config.History.EndTime);
                 if (parsed == null) return $"Invalid history end time: {config.History.EndTime}";
             }
+            if (config.Source.SamplingInterval != null)
+            {
+                log.LogWarning("source.sampling-interval is deprecated. Use subscriptions.sampling-interval instead.");
+                config.Subscriptions.SamplingInterval = config.Source.SamplingInterval.Value;
```
toondaey, line 85: What happens if the same property is somehow, also set in the `Subscriptions` config section? Which one takes precedence if it, hypothetically, happens?

ExtractorLauncher/ExtractorStarter.cs
```diff
@@ -79,6 +79,16 @@ public static class ExtractorStarter
...
+            }
+            if (config.Source.QueueLength != null)
+            {
+                log.LogWarning("source.queue-length is deprecated. Use subscriptions.queue-length instead.");
+                config.Subscriptions.QueueLength = config.Source.QueueLength.Value;
```
toondaey, line 90: Same here.

toondaey (changes requested):

ExtractorLauncher/ExtractorStarter.cs
einarmo: The source does, due to this. However it is hard to detect whether the value is set in Subscriptions, since it has a default value there.

This is fine. If somebody _does_ set it in `source`, it should mean that they are using old config, and that they meant to set it.

toondaey: Oh right! Makes sense now.

toondaey (approved):

PR #372, opened 2023-03-16
Improve handling of incorrect proto node ids

einarmo:
We were silently converting them to other types, which is just a bad idea in general. This instead logs a warning.


toondaey (approved):
lgtm

PR #380, opened 2023-04-04
Add certificate auth

mathialo:
Update extractor utils and add new version to manifest


einarmo (commented):
The test project contains a reference to `Cognite.ExtractorUtils.Testing`. It won't break in this case, but to avoid having a transient dependency on a different version, perhaps update that one too?

einarmo (approved):

PR #383, opened 2023-04-17
Fix reconnect after service level change

einarmo:
The extractor uses the stored EndpointUrl to check for whether the endpoint has changed after an attempted redundant reconnect. This didn't work properly, because it was set inside CreateSessionDirect.

This removes that, which should fix the problem.


toondaey (approved):
👍🏾

PR #390, opened 2023-04-26
Add data-set config object from utils

einarmo:
Aligns with other extractors.

### DIFF
--- a/Extractor/Pushers/CDFPusher.cs
+++ b/Extractor/Pushers/CDFPusher.cs
@@ -230,7 +230,7 @@ namespace Cognite.OpcUa.Pushers
             try
             {
                 var result = await destination.EnsureEventsExistsAsync(eventList
-                    .Select(evt => evt.ToCDFEvent(Extractor, config.DataSetId, nodeToAssetIds))
+                    .Select(evt => evt.ToCDFEvent(Extractor, config.DataSet?.Id, nodeToAssetIds))
                     .Where(evt => evt != null), RetryMode.OnError, SanitationMode.Clean, token);
 
                 log.LogResult(result, RequestType.CreateEvents, false, LogLevel.Debug);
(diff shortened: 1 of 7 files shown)


Extractor/Pushers/CDFPusher.cs
```diff
@@ -1253,18 +1253,18 @@ private async Task MarkReferencesAsDeleted(IEnumerable<string> externalIds, Canc
         /// </summary>
         private async Task EnsureConfigInit(CancellationToken token)
         {
-            if (!config.DataSetId.HasValue && !string.IsNullOrEmpty(config.DataSetExternalId))
+            if (config.DataSet != null)
```
ozangoktan, line 1256: The old if block was checking if dataset was defined by externalid, and retrieving the numeric id if necessary. 

This one is checking the top level object, which is also checked in DataSets.GetId method. Should we rework this section a bit?

Extractor/Pushers/CDFPusher.cs
```diff
@@ -1253,18 +1253,18 @@ private async Task MarkReferencesAsDeleted(IEnumerable<string> externalIds, Canc
...
-                    config.DataSetId = dataSet.First().Id;
+                    var id = await destination.CogniteClient.DataSets.GetId(config.DataSet, token);
+                    config.DataSet.Id = id;
                 }
                 catch (ResponseException ex)
                 {
                     log.LogError("Could not fetch data set by external id. It may not exist, or the user may lack" +
                         " sufficient access rights. Project {Project} at {Host}, id {Id}: {Message}",
-                        config.Project, config.Host, config.DataSetExternalId, ex.Message);
+                        config.Project, config.Host, config.DataSet.ExternalId, ex.Message);
```
ozangoktan, line 1267:
```suggestion
                        config.Project, config.Host, config.DataSet?.ExternalId, ex.Message);
```
This one might be unnecessary if this section is reworked.

Extractor/Pushers/CDFPusher.cs
einarmo: It's the same, the GetId method checks if ExternalId is defined.

ozangoktan (approved):
LGTM

PR #393, opened 2023-04-27
Session manager improvement

einarmo:


Server/Program.cs
```diff
@@ -183,6 +183,34 @@ void ServerUpdate(object state)
                 exitEvent.Set();
             };
 
+            var _ = Task.Run(() =>
```
ozangoktan, line 186: Is this a new feature for setting service level via console?

Extractor/SessionManager.cs
```diff
@@ -544,7 +525,7 @@ private async Task UpdateServiceLevelInner(byte newLevel, bool fromConnectionCha
                 // We don't want to reconnect too frequently
                 if (lastLowSLConnectAttempt != null)
                 {
-                    var timeSinceLastAttempt = DateTime.UtcNow - lastLowSLConnectAttempt.Value;
+                    var timeSinceLastAttempt = (DateTime.UtcNow - lastLowSLConnectAttempt.Value) + TimeSpan.FromSeconds(5);
```
ozangoktan, line 528: Would this not have the effect of internally reducing config.Redundancy.ReconnectIntervalValue.Value by 5 seconds?

mathialo (approved):

PR #397, opened 2023-05-11
Target Flexible Data Models

einarmo:
Initial version of feature writing to flexible data models through DMSv3.

What this does is create a view and container for each referenced OPC-UA type, then map instances to those types, in the simplest possible terms.

It will likely need to evolve a great deal in the future. This is not made available to users yet.


Extractor/NodeSources/NodeSetSource.cs
```diff
@@ -136,7 +142,7 @@ private void LoadReferences()
                     {
                         references[targetId] = targetRefs = new Dictionary<(NodeId, NodeId, bool), IReference>();
                     }
-                    if (!targetRefs.ContainsKey((node.NodeId, reference.ReferenceTypeId, reference.IsInverse)))
+                    if (!targetRefs.ContainsKey((node.NodeId, reference.ReferenceTypeId, !reference.IsInverse)))
```
ozangoktan, line 145: The flag here seems to have flipped, but I could not see an obvious change upstream of where it comes from. Raising it to confirm.

Extractor/NodeSources/NodeSetSource.cs
```diff
@@ -362,7 +412,7 @@ public void BuildNodes(IEnumerable<NodeId> rootNodes, bool isFullBrowse)
                 isFullBrowse);
         }
 
-        private void GetRelationshipData()
+        private void GetRelationshipData(bool getPropertyReferences, HashSet<NodeId> additionalKnownNodes, List<(NodeId, IReference)> newNodes)
```
ozangoktan, line 415: Minor: Method name starts with get despite returning void, and no obvious out parameters

Extractor/NodeSources/NodeSetSource.cs
```diff
@@ -330,9 +335,54 @@ public void BuildNodes(IEnumerable<NodeId> rootNodes, bool isFullBrowse)
                 InitNodeState(update, node);
             }
 
+        }
+
+        public override async Task<NodeSourceResult?> ParseResults(CancellationToken token)
+        {
+            if (!NodeMap.Any()) return null;
+
+            FinalDestinationObjects.Clear();
```
ozangoktan, line 344: Should we implement locks on these objects? They are referenced from a lot of places.

Extractor/Nodes/BaseUANode.cs
```diff
@@ -152,6 +152,7 @@ public IEnumerable<BaseUANode> GetAllProperties()
 
         public BaseUANode(NodeId id, BaseUANode? parent, NodeId? parentId)
         {
+            // if (id.IsNullNodeId) throw new ArgumentException("Id may not be the null NodeId", nameof(id));
```
ozangoktan, line 155:
```suggestion
```

ConfigurationTool/Checks/Events.cs
```diff
@@ -63,24 +63,37 @@ public async Task GetEventConfig(CancellationToken token)
...
-
-            var emittedEvents = referencedEvents.ToList();
+            var emittedEvents = emitterReferences.Select(evt => evt.Id).ToList();
 
             if (emittedEvents.Any())
             {
                 log.LogInformation("Identified {Count} events by looking at GeneratesEvent references", emittedEvents.Count);
+                foreach (var rf in emitterReferences)
+                {
+                    log.LogInformation("Test {T} {E}", rf.ParentId, rf.Id);
```
ozangoktan, line 95: Maybe this should be moved to a more verbose log level?

Extractor/Pushers/CDFPusher.cs
```diff
@@ -301,7 +319,14 @@ public sealed class CDFPusher : IPusher
 
             log.LogInformation("Testing {TotalNodesToTest} nodes against CDF", variables.Count() + objects.Count());
 
-            if (fullConfig.DryRun) return result;
+            if (fullConfig.DryRun)
+            {
+                if (fdmDestination != null)
+                {
+                    await fdmDestination.PushNodes(objects, variables, references, Extractor, token);
```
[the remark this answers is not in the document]
einarmo: Yeah, it's _very_ nice to get a dump of all the types and such when doing a dry run. Without this the dry run essentially doesn't show the user what we will make in FDM.

Extractor/Types/UAPropertySerializer.cs
```diff
@@ -31,6 +31,20 @@
 
 namespace Cognite.OpcUa.Types
 {
+    public enum StringConverterMode
+    {
+        Simple,
+        Json,
+        ReversibleJson
```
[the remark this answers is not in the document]
einarmo: So for serializing most OPC-UA types we use the JSON encoder from the OPC-UA SDK, which is somewhat finnicky, it wasn't really intended to be used this way. Reversible JSON is telling it to include type information in the payload, which is handy if we're going to convert it back to OPC-UA later.

Extractor/NodeSources/NodeSetSource.cs
einarmo: It was a bug, actually. This feature never worked as it should... It is pretty niche.

Extractor/NodeSources/NodeSetSource.cs
einarmo: They are protected, and I don't see any parallelism in this class, it shouldn't be necessary?

ConfigurationTool/Checks/Events.cs
einarmo: Should probably just be removed...

Extractor/NodeSources/NodeSetSource.cs
ozangoktan: Should be fine then 👍🏼

Extractor/Pushers/CDFPusher.cs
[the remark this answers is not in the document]
ozangoktan: Though I see the use case, would argue that it invalidates dry run. Maybe a config option could be put in place? 

**_--dry-run-with-type-schemas_** or something similar?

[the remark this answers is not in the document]
einarmo: Oh. It doesn't actually push, it's still a dry run.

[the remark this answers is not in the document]
ozangoktan: Ah, it does that internally. 🚀

Extractor/Types/UAPropertySerializer.cs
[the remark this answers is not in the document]
einarmo: This produces a very different type of JSON, I think having it abstracted from the outside is fine. The values produced by this contains some type information in certain cases.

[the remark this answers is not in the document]
ozangoktan: Ah, looks like this is an OPCUA thing -> https://reference.opcfoundation.org/Core/Part6/v105/docs/5.4

👍🏼

ozangoktan (approved):

PR #405, opened 2023-05-24
Refactoring

ozangoktan:
While reviewing the FDM targeting branch, I wanted to do a bit of refactoring. Renamed some classes and (hopefully) streamlined class inheritance.

### DIFF
--- a/Extractor/Pushers/FDM/Types/ReferenceTypes.cs
+++ b/Extractor/Pushers/FDM/Types/ReferenceTypes.cs
@@ -5,9 +5,9 @@ using Opc.Ua;
 
 namespace Cognite.OpcUa.Pushers.FDM.Types
 {
-    public class ReferenceNodeType : ReferenceNode
+    public class NodeTypeReference : ReferenceNode
     {
-        public ReferenceNodeType(NodeClass nodeClass, string browseName, string externalId, UAReference uaReference)
+        public NodeTypeReference(NodeClass nodeClass, string browseName, string externalId, UAReference uaReference)
             : base(nodeClass, browseName, externalId, uaReference)
         {
         }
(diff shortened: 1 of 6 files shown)


Extractor/Pushers/FDM/Types/ReferenceTypes.cs
```diff
@@ -0,0 +1,44 @@
+using Cognite.OpcUa.Nodes;
+using Cognite.OpcUa.Types;
+using CogniteSdk.Beta.DataModels;
+using Opc.Ua;
+
+namespace Cognite.OpcUa.Pushers.FDM.Types
+{
+    public class ReferenceNodeType : ReferenceNode
```
einarmo, line 8: So this renaming doesn't really make sense. A "Reference Node Type" is, presumably, the type of a "Reference Node", whatever that is, which is not what this is. I used "NodeTypeReference" because it is a reference from a Node Type.

Extractor/TypeCollectors/TypeManager.cs
```diff
@@ -322,106 +323,49 @@ private IEnumerable<TypeField> CollectNormalNode(NodeId id)
         #endregion
 
         #region typeGetters
-        public UADataType GetDataType(NodeId nodeId)
+        private T GetType<T>(NodeId nodeId, Func<NodeId, T> constructor, string typeName) where T : BaseUAType
```
einarmo, line 326: Neat. That's a good way to do it without a runtime cost.

einarmo (commented):
Mostly good changes, though I'm a little sceptical of the name changes. They may seem more natural, but that doesn't help if they are incorrect. We may need to think more about the names.

ozangoktan:
> Mostly good changes, though I'm a little sceptical of the name changes. They may seem more natural, but that doesn't help if they are incorrect. We may need to think more about the names.

Happy to make further changes to naming if you have suggestions, though we shouldn't have to exactly match the domain names. We can have some annotations for that at the worst case.

einarmo (approved):
We can improve further on naming later

PR #415, opened 2023-06-01
Add otpions for LifetimeCount and KeepAliveCount

einarmo:
Also delete subscriptions on session shutdown.

AkerBP are having some issues with sessions, and prediktor seemed to think setting these may help. They are fine to expose to users anyway.


ozangoktan (approved):
LGTM, DMSv3 changes are internal so my previous comment does not apply.

PR #418, opened 2023-06-06
Add extra metadata per type

einarmo:
This is necessary for the interface, it turns out. We _could_ try to infer this information from the node hierarchy, but this is probably a better solution.


ozangoktan (approved):
LGTM!

PR #425, opened 2023-06-23
Add NodeClass and Mandatory to type metadata

einarmo:


Extractor/Pushers/FDM/TypeMetadata.cs
```diff
@@ -13,6 +13,11 @@ public class PropertyNode
         public int[]? ArrayDimensions { get; set; }
         public int? ValueRank { get; set; }
         public string? BrowseName { get; set; }
+        public int? NodeClass { get; set; }
+        public bool Mandatory { get; set; }
```
toondaey, line 17: We could probably use `IsMandatory` as a more descriptive boolean property.

toondaey (approved):
👍🏾

PR #435, opened 2023-06-28
Initial preparation for supporting writing to any/all cdf targets

toondaey:


ExtractorLauncher/ExtractorStarter.cs
```diff
@@ -124,6 +125,47 @@ public static class ExtractorStarter
...
+                    RawMetadata = new RawMetadataTargetConfig
+                    {
+                        Database = rawMetadata?.Database,
+                        AssetsTable = rawMetadata?.AssetsTable,
+                        TimeseriesTable = rawMetadata?.TimeseriesTable,
+                        RelationshipsTable = rawMetadata?.RelationshipsTable
+                    }
+                };
+            }
+            if (config.Cognite?.FlexibleDataModels != null)
```
einarmo, line 158: You don't need this mapping, because this isn't public.

schema/cognite_config.schema.json
```diff
@@ -57,6 +59,100 @@
                 }
             }
         },
```
einarmo, line 61: Remove these blocks entirely.

ExtractorLauncher/ExtractorStarter.cs
```diff
@@ -124,6 +125,47 @@ public static class ExtractorStarter
...
+                    },
+                    RawMetadata = new RawMetadataTargetConfig
+                    {
+                        Database = rawMetadata?.Database,
+                        AssetsTable = rawMetadata?.AssetsTable,
+                        TimeseriesTable = rawMetadata?.TimeseriesTable,
+                        RelationshipsTable = rawMetadata?.RelationshipsTable
+                    }
+                };
+            }
```
einarmo, line 157:
```c#
else if (config.Cognite.MetadataTargets == null) {
    if (skipMetadata) {
        Deprecation warning
    }
    else {
         Enable clean by default + deprecation warning Default writing to clean is deprecated, in the future not setting a metadata target will not write metadata to CDF at all
    }
}
```

Extractor/Config/CogniteConfig.cs
```diff
@@ -136,6 +145,34 @@ public class RawMetadataConfig
...
+        public string? Database { get; set; }
+        public string? AssetsTable { get; set; }
+        public string? TimeseriesTable { get; set; }
+        public string? RelationshipsTable { get; set; }
+    }
+    public class CleanMetadataTargetConfig
+    {
+        public bool Assets { get; set; } = true;
+        public bool Timeseries { get; set; } = true;
+        public bool Relationships { get; set; } = true;
```
einarmo, line 174: These shouldn't be true by default

Extractor/Config/CogniteConfig.cs
```diff
@@ -136,6 +145,34 @@ public class RawMetadataConfig
...
+    public class MetadataTargetsConfig 
+    {
+        /// <summary>
+        /// Raw metadata targets config
+        /// </summary>
+        public RawMetadataTargetConfig? RawMetadata { get; set; }
+        /// <summary>
+        /// Clean metadata targets config
+        /// </summary>
+        public CleanMetadataTargetConfig? CleanMetadata { get; set; }
```
einarmo, line 157: Just call this `clean`, and `raw`

Extractor/Config/CogniteConfig.cs
```diff
@@ -136,6 +145,34 @@ public class RawMetadataConfig
...
+        /// <summary>
+        /// FDM destination config
+        /// </summary>
+        public FdmDestinationConfig? FlexibleDataModels { get; set; }
```
einarmo, line 161: Should be data-models

ExtractorLauncher/ExtractorStarter.cs
einarmo: Also if this is the case log

`Default writing to clean is deprecated, in the future not setting a metadata target will not write metadata to CDF at all`

einarmo (changes requested):

schema/cognite_config.schema.json
```diff
@@ -29,31 +29,97 @@
...
-            "required": [ "database" ],
             "properties": {
-                "database": {
-                    "type": "string",
-                    "description": "Raw database"
-                },
-                "assets-table": {
-                    "type": "string",
-                    "description": "Raw table to use for assets"
+                "raw-metadata": {
```
einarmo, line 37: You changed the name of these in config, need to update here too.

einarmo: Also, I think you need to update the example config file.

einarmo (approved):

PR #444, opened 2023-07-12
release 2.20.1

ozangoktan:


ozangoktan:
> Seems this is released already. https://github.com/cognitedata/opcua-extractor-net/releases/tag/2.20.1

That is a failed release with regards to uploading to CDF, see -> https://github.com/cognitedata/opcua-extractor-net/actions/runs/5532266762/jobs/10094243329

PR #448, opened 2023-07-24
Fix FileNotFound issue when pushing datapoints on windows

einarmo:
It's some bizarre dependency error. Makes no damn sense, probably something really weird with how the protobuf library is loaded.


ozangoktan (approved):
LGTM! 🚀

PR #461, opened 2023-08-04
Add metadata about properties to FDM nodes

einarmo:
We need this because there is no way to distinguish between a property being null and not existing in FDM. The map is to the original NodeId of the property, which will be useful if we ever need to map those over.


ozangoktan (approved):
LGTM

PR #492, opened 2023-09-20
Retry more

einarmo:
Add a pile of status codes we can reasonably retry, and make it possible to configure status codes using strings and hexadecimal.


rsjr (approved):
aprob

PR #493, opened 2023-09-21
Update to codesign v2

einarmo:
Done in the same style as db/opc-classic, so we release differently now and build the installer in CI.


rsjr (approved):
aprob

PR #500, opened 2023-10-04
Centralize subscription management

einarmo:
This moves all logic for creating and maintaining subscriptions out of UAClient and into a dedicated subscription manager. This tidies up a few different features related to subscriptions that were interacting dangerously.

This is a bit less convenient for testing, but it does tidy up UAClient a bit, which is good, since that class is too big.


Extractor/Config/SourceConfig.cs
```diff
@@ -315,6 +315,38 @@ public HashSet<uint> FinalRetryStatusCodes
...
+                if (silentExc.InnerServiceException != null)
+                {
+                    return statusCodes.Contains(silentExc.InnerServiceException.StatusCode);
+                }
+            }
+            else if (ex is AggregateException aex)
+            {
+                // Only retry aggregate exceptions if one of the inner exceptions should be retried...
+                var flat = aex.Flatten();
+                return aex.InnerExceptions.Any(e => ShouldRetryException(e, statusCodes));
```
ozangoktan, line 341: Should we not use the flattened exception here? Or remove it?

ConfigurationTool/ConfigToolRuntime.cs
```diff
@@ -43,6 +43,13 @@ public ConfigToolRuntime(IServiceProvider provider, FullConfig config, FullConfi
             this.baseConfig = baseConfig;
             this.output = output;
             log = provider.GetRequiredService<ILogger<UAServerExplorer>>();
+
+            // Set retry config lower here, otherwise this might take a really, really long time.
+            // We do a few retries, just to avoid the worst flakyness on the server.
+            config.Source.Retries = new UARetryConfig
```
ozangoktan, line 49: Would be good to externally document that the configuration tool ignores the set value for retries.

Extractor/Config/SourceConfig.cs
einarmo: Yep, typo.

ConfigurationTool/ConfigToolRuntime.cs
einarmo: Yeah, we should add that to the docs page on the config tool.

ozangoktan: I think we should also have some indicator in the following places:
- https://github.com/cognitedata/opcua-extractor-net/blob/f1ee3bd8fe26f1bfeb21345cc4b6379301bdb655/Extractor/Config/SourceConfig.cs#L202
- https://github.com/cognitedata/opcua-extractor-net/blob/f1ee3bd8fe26f1bfeb21345cc4b6379301bdb655/config/config.example.yml#L134

einarmo: Sure.

ozangoktan (approved):
LGTM 🎸

PR #502, opened 2023-10-09
Improve logging in node sources

einarmo:
I wasn't able to reproduce the bug (DOG-2235), and I won't be able to without better logging, so this just improves logging in a few places. Part of this is due to a small regression caused by the node sources change. We still want to communicate the context in which data is being extracted, it is really important for ensuring that users understand what the extractor is doing.


ozangoktan (approved):
LGTM

PR #504, opened 2023-10-16
Add cache for subscription states

einarmo:
This adds a cache containing the state of each subscription, and when it was last updated. It is effectively a very overkill solution for making subscription recreation wait a bit before they run, to help prevent subscriptions from being queued for deletion, then recover after being fully created.

The solution is a bit much, but it's relatively tidy, and it should work. This is pretty much impossible to test.

### DIFF
--- a/Test/Utils/BaseExtractorTestFixture.cs
+++ b/Test/Utils/BaseExtractorTestFixture.cs
@@ -170,7 +170,6 @@ namespace Test.Utils
             var newServices = new ServiceCollection();
             foreach (var service in Services)
             {
-
                 newServices.Add(service);
             }
             CommonTestUtils.AddDummyProvider("test", CDFMockHandler.MockMode.None, true, newServices);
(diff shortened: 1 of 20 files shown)


Test/Utils/BaseExtractorTestFixture.cs
```diff
@@ -167,8 +168,9 @@ public async Task ClearLiteDB(InfluxDBClient client)
         public (CDFMockHandler, CDFPusher) GetCDFPusher()
         {
             var newServices = new ServiceCollection();
-            foreach (var service in Services) {
-                
+            foreach (var service in Services)
+            {
+
```
ozangoktan, line 173:
```suggestion
```

ozangoktan (approved):
LGTM

Test/Utils/BaseExtractorTestFixture.cs
```diff
@@ -167,8 +168,9 @@ public async Task ClearLiteDB(InfluxDBClient client)
...
                 newServices.Add(service);
```
ozangoktan, line 174:
```suggestion
            foreach (var service in Services)
            {
                newServices.Add(service);
```

PR #508, opened 2023-10-17
Separate session context from session

einarmo:
This is a bit tricky, but it might allow the extractor to keep working even when connection to the server is lost. It should generally be fine to continue going with the _previous_ session context. If the setup is such that this won't work, the extractor should be configured to restart hard on connection loss. Most servers will work fine with this.

### DIFF
--- a/Extractor/Nodes/UAVariable.cs
+++ b/Extractor/Nodes/UAVariable.cs
@@ -290,7 +290,7 @@ namespace Cognite.OpcUa.Nodes
             return (Id, -1);
         }
 
-        public override Dictionary<string, string>? GetExtraMetadata(FullConfig config, IUAClientAccess client)
+        public override Dictionary<string, string>? GetExtraMetadata(FullConfig config, SessionContext context, StringConverter converter)
         {
             Dictionary<string, string>? fields = null;
             if (config.Extraction.NodeTypes.Metadata && FullAttributes.TypeDefinition?.Name != null)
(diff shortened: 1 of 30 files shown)


Extractor/UAClient.cs
```diff
@@ -234,18 +231,9 @@ private async Task StartSession(int timeout = -1)
 
             // A restarted Session might mean a restarted server, so all server-relevant data must be cleared.
             // This includes any stored NodeId, which may refer to an outdated namespaceIndex
```
ozangoktan, line 233: This comment does not make too much sense with the LoadAppConfig call

ozangoktan (commented):
Generally looks good! Had a couple of comments.

Extractor/Nodes/UAVariable.cs
```diff
@@ -532,9 +532,11 @@ public UAVariableMember(UAVariable parent, int index)
             TSParent = parent;
         }
 
-        public override string? GetUniqueId(IUAClientAccess client)
+        public override string? GetUniqueId(SessionContext? context)
         {
-            return client.GetUniqueId(Id, Index);
+            if (context == null) throw new InvalidOperationException("Attempt to get unique ID without initialized client");
```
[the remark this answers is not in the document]
einarmo: Missed this one, context should never be nullable anymore.

Extractor/UAClient.cs
einarmo: No... I'll remove it.

ozangoktan (approved):
LGTM!

PR #510, opened 2023-10-23
Track namespacepublicationdate - dog-2234

toondaey:


Extractor/RebrowseTriggerManager.cs
```diff
@@ -104,48 +132,181 @@ public async Task EnableCustomServerSubscriptions(CancellationToken token)
 
                 foreach (var reference in references)
                 {
-                    if (!targetNodes.Contains(reference.DisplayName.Text)) continue;
+                    if (!targetNodes.Contains(reference.DisplayName.Text))
+                        continue;
                     var id = _uaClient.ToNodeId(reference.NodeId);
                     nodes.TryAdd(id, (id, reference.DisplayName.Text));
+                    _extractionStates.TryAdd(
+                        id.ToString(),
```
einarmo, line 140: Avoid using the string representation of nodes. Use `GetUniqueId`.

Extractor/RebrowseTriggerManager.cs
```diff
@@ -104,48 +132,181 @@ public async Task EnableCustomServerSubscriptions(CancellationToken token)
...
+                        new NamespacePublicationDateState(id.ToString())
```
einarmo, line 141: Same here

Test/Integration/RebrowseTriggerManagerTests.cs
```diff
@@ -66,76 +75,143 @@ public async Task TestRebrowseIsNotSubscribed(RebrowseTriggersConfig config)
...
                 Targets = new RebrowseTriggerTargets { NamespacePublicationDate = true }
             };
-            using var extractor = tester.BuildExtractor(true, pushers: pusher);
-
-
-            // Act
+            tester.Config.StateStorage = new StateStorageConfig
+            {
+                Database = StateStoreConfig.StorageType.LiteDb,
+                Location = "lite-state-rebrowse.db",
```
einarmo, line 86: We do this in some other tests, delete the state store file in a try block when starting the test.

Extractor/RebrowseTriggerManager.cs
```diff
@@ -104,48 +132,181 @@ public async Task EnableCustomServerSubscriptions(CancellationToken token)
...
+                throw new InvalidOperationException("Client not initialized");
 
-            _uaClient.SubscriptionManager.EnqueueTask(new RebrowseTriggerSubscriptionTask(
-                OnRebrowseTriggerNotification, nodes));
+            _uaClient.SubscriptionManager.EnqueueTask(
+                new RebrowseTriggerSubscriptionTask(OnRebrowseTriggerNotification(token), nodes)
+            );
         }
+
+        private void OnServerReconnect() { }
```
einarmo, line 310: What is this for? It appears to be unused.

einarmo (commented):
This looks decent. I have some nit, and we do need to prevent double browse on startup, if at all possible.

einarmo (approved):

PR #511, opened 2023-10-23
Extract various extraction toggles

einarmo:
Makes for more readable code, and improves the distinction between FDM (in the pusher) and loading from OPC-UA, even if the two are somewhat related.

### DIFF
--- a/Extractor/Config/ConfigToggles.cs
+++ b/Extractor/Config/ConfigToggles.cs
@@ -47,7 +47,7 @@ namespace Cognite.OpcUa.Config
         /// </summary>
         public bool LoadReferenceTypes => FdmEnabled || config.Extraction.Relationships.Enabled;
         /// <summary>
-        /// Should the type manager load object- and variable types when reading the type hierarchy?
+        /// Should the type manager load object and variable types when reading the type hierarchy?
         /// </summary>
         public bool LoadTypeDefinitions => FdmEnabled;
         /// <summary>
--- a/Extractor/Config/ConfigToggles.cs
+++ b/Extractor/Config/ConfigToggles.cs
@@ -1,0 +1,58 @@
+namespace Cognite.OpcUa.Config
+{
+    /// <summary>
+    /// Class wrapping a config instance, enabling features that may be enabled from different places.
+    /// </summary>
+    public class ConfigToggles
+    {
+        private readonly FullConfig config;
+        public ConfigToggles(FullConfig config)
+        {
+            this.config = config;
+        }
+
+        private bool FdmEnabled => config.Cognite?.MetadataTargets?.DataModels?.Enabled ?? false;
+
+        /// <summary>
+        /// Should the extractor load references to and from properties?
+        /// </summary>
+        public bool MapPropertyReferences => FdmEnabled;
+
+        /// <summary>
+        /// Should the extractor get references to types when loading non-hierarchical references?
+        /// </summary>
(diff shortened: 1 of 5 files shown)


Extractor/Config/ConfigToggles.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// <summary>
+        /// Should the extractor get references to types when loading non-hierarchical references?
+        /// </summary>
+        public bool GetNonHierarchicalTypeReferences => FdmEnabled || config.Extraction.NodeTypes.AsNodes;
+        /// <summary>
+        /// Should the extractor create unmapped nodes it encounters when reading non-hierarchical references, or just ignore them?
+        /// </summary>
+        public bool CreateUnknownReferencedNodes => FdmEnabled || config.Extraction.Relationships.CreateReferencedNodes;
+        /// <summary>
+        /// Should the extractor create nodes with NodeClass ObjectType, ReferenceType, VariableType, or DataType?
```
ozangoktan, line 30: This sounds like the bool makes a distinction between 4 different cases, can we improve the comment here?

Extractor/Config/ConfigToggles.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        // Type manager
+        /// <summary>
+        /// Should the type manager load data types when reading the type hierarchy?
+        /// </summary>
+        public bool LoadDataTypes => FdmEnabled || config.Extraction.DataTypes.AutoIdentifyTypes;
+        /// <summary>
+        /// Should the type manager load event types when reading the type hierarchy?
+        /// Since event types are also type definitions, this is enabled by default if
+        /// LoadTypeDefinitions is true.
```
ozangoktan, line 42: This refers to some external logic. Perhaps we should exclude this information from here

Extractor/Config/ConfigToggles.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// </summary>
+        public bool LoadEventTypes => config.Events.Enabled;
+        /// <summary>
+        /// Should the type manager load reference types when reading the type hierarchy?
+        /// </summary>
+        public bool LoadReferenceTypes => FdmEnabled || config.Extraction.Relationships.Enabled;
+        /// <summary>
+        /// Should the type manager load object- and variable types when reading the type hierarchy?
```
ozangoktan, line 50:
```suggestion
        /// Should the type manager load object and variable types when reading the type hierarchy?
```

Extractor/Config/ConfigToggles.cs
einarmo: Sure

Extractor/Config/ConfigToggles.cs
einarmo: They are toggles, each enable a feature independently.

Extractor/Config/ConfigToggles.cs
einarmo: Actually, this doesn't refer to external logic. It refers to the structure of the base OPC-UA type hierarchy, so it is inherent to OPC-UA.

The type definition hierarchy looks like this:

 - ObjectTypes
   - EventTypes
 - VariableTypes

so if type definitions are enabled, we will get event types as well.

Extractor/Config/ConfigToggles.cs
einarmo: I improved the docs a bit

ozangoktan (approved):
LGTM 🔥

PR #513, opened 2023-10-25
Separate model and instance spaces, and make version configurable

einarmo:
A sanity check is much appreciated here, it's pretty important that the correct space is used. Instances and references
(DirectRelationIdentifier) should all point to InstanceSpace. References to containers, views, and properties should all use ModelSpace.


toondaey (approved):
We could maybe use some switch expressions in the `DMSValueConverter` otherwise, lgtm!

PR #516, opened 2023-10-26
Slightly improve error handling when creating subscriptions

einarmo:
Move a few more retries out, and make the inner subscription loop wait for a valid session.


ozangoktan (approved):
LGTM

PR #517, opened 2023-10-27
Fdm deletes

einarmo:
Ended up having to fetch the edges to delete based on the nodes, and store additional information in state store about the nodes and edges to be deleted.


ozangoktan (approved):
LGTM

PR #526, opened 2023-11-02
Clear delete marker on clean resources

einarmo:


Extractor/Pushers/PusherUtils.cs
```diff
@@ -136,11 +136,13 @@ public static long GetTimestampValue(object value)
                         out _);
 
                 if (newMetaData.Any() && (old.Metadata == null
-                        || !newMetaData.All(kvp => old.Metadata.TryGetValue(kvp.Key, out var oldVal) && kvp.Value == oldVal)))
+                        || !newMetaData.All(kvp => old.Metadata.TryGetValue(kvp.Key, out var oldVal) && kvp.Value == oldVal))
+                    || config.Extraction.Deletes.Enabled && old.Metadata.ContainsKey(config.Extraction.Deletes.DeleteMarker))
```
ozangoktan, line 140: Might be a good idea to refactor this block into its own function or something else to make it less complex.

Extractor/Pushers/PusherUtils.cs
```diff
@@ -191,7 +193,8 @@ public static long GetTimestampValue(object value)
...
```
ozangoktan, line 197: Same as above

ozangoktan (approved):
LGTM 🚀

PR #528, opened 2023-11-06
Increase slack and add option to disable sub recreation

einarmo:
Might help, might not. Probably doesn't hurt.

### DIFF
--- a/Extractor/Subscriptions/RecreateSubscriptionTask.cs
+++ b/Extractor/Subscriptions/RecreateSubscriptionTask.cs
@@ -1,5 +1,6 @@
 ﻿using Cognite.OpcUa.Config;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.ObjectPool;
 using Opc.Ua;
 using Opc.Ua.Client;
 using System;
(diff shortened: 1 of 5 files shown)


Extractor/Subscriptions/RecreateSubscriptionTask.cs
```diff
@@ -65,24 +66,30 @@ public override async Task Run(ILogger logger, SessionManager sessionManager, Fu
...
             {
-                logger.LogWarning("Server is available, but subscription is not responding to notifications. Attempting to recreate.");
+                logger.LogWarning("Server is available, but subscription {Name} is not responding to notifications. Attempting to recreate.", SubscriptionName);
                 await session.RemoveSubscriptionAsync(oldSubscription);
             }
             catch (ServiceResultException serviceEx)
             {
                 var symId = StatusCode.LookupSymbolicId(serviceEx.StatusCode);
-                logger.LogWarning("Error attempting to remove subscription from the server: {Err}. It has most likely been dropped. Attempting to recreate...", symId);
+                logger.LogWarning("Error attempting to remove subscription {Name} from the server: {Err}. It has most likely been dropped. Attempting to recreate...", symId, SubscriptionName);
```
ozangoktan, line 92: Nice!

ozangoktan (approved):
LGTM!

Extractor/Subscriptions/RecreateSubscriptionTask.cs
einarmo: Good that you commented there, actually, it's wrong, the two logging arguments are in the wrong order.

PR #534, opened 2023-11-09
Improve check for whether we should update views or not

einarmo:
This is more generic, and less prone to weirdness. It isn't perfect, but it should be sufficient for the extractor.


toondaey (approved):
👍🏾

PR #535, opened 2023-11-09
test: attempt to fix failing test

toondaey:


Extractor/RebrowseTriggerManager.cs
```diff
@@ -186,26 +187,23 @@ CancellationToken token
                 .Select(
                     (dv, i) =>
                         (
-                            nodes[i].Item1,
+                            _uaClient.GetUniqueId(nodes[i].Item1),
                             dv.GetValue<DateTime>(default).ToUnixTimeMilliseconds()
                         )
                 )
                 .ToDictionary(item => item.Item1, item => item.Item2);
```
einarmo, line 194: Use `ToDictionary` directly instead of going through a tuple.

Extractor/RebrowseTriggerManager.cs
```diff
@@ -220,16 +218,15 @@ CancellationToken token
                 try
                 {
                     var values = item.DequeueValues();
-                    var valueTime = values[0]
-                        .GetValue<DateTime>(default)
-                        .ToUnixTimeMilliseconds();
-                    var id = item.ResolvedNodeId.ToString();
+                    var valueTime = values[0].GetValue<DateTime>(default).ToUnixTimeMilliseconds();
```
einarmo, line 221: I am pretty sure this will fail. Try this: `DateTime.MinValue.ToUnixTimeMilliseconds()`. I think that will throw an exception.

einarmo (commented):
I have some comments

Extractor/RebrowseTriggerManager.cs
toondaey: This may be necessary because there's no way to get the index in `ToDicationary`.

einarmo: Could you use `Zip`?

Extractor/RebrowseTriggerManager.cs
einarmo: Sorry, I wasn't being clear. `default(DateTime) == DateTime.MinValue`, but `DateTime.MinValue.ToUnixTimeMilliseconds()` will fail, so `default(DateTime).ToUnixTimeMilliseconds()` will also fail.

Extractor/RebrowseTriggerManager.cs
toondaey: With `Zip` we'd have to create anonymous objects that can be used in `ToDictionary`.

einarmo (approved):

PR #540, opened 2023-11-16
Better aggregation of history errors

einarmo:
The logs get insane if there are thousands of failures here, it would be much more useful to just count the number of exceptions of each type.


ozangoktan (approved):
LGTM 💯

PR #545, opened 2023-11-27
Try to fix subscription issue

einarmo:
Not able to reproduce this, this fix might help, or it might do nothing, but it probably won't cause problems.


ozangoktan (approved):
LGTM!

PR #574, opened 2024-01-09
Restart history when subscriptions are recreated

einarmo:
This is a fairly complex change. The issue with this new feature (title) is that we are already triggering restart of history from multiple different places. This already isn't ideal, since we only really want to run history once in parallel. (At least for the same set of nodes). This moves the logic for handling whether a change in extractor status should lead to a restart of history into HistoryReader, and has everything else just calling simple methods to update the state of the history reader.

Hopefully this works, and doesn't just cause more trouble.


ozangoktan (approved):
LGTM!

PR #595, opened 2024-02-01
Support for status codes

einarmo:


Extractor/History/HistoryScheduler.cs
```diff
@@ -94,44 +95,45 @@ public class HistoryScheduler : SharedResourceScheduler<HistoryReadNode>
...
+                    ? config.History.DataNodesChunk
+                    : config.History.EventNodesChunk,
                   resource,
                   token)
         {
             this.log = log;
             this.uaClient = uaClient;
             this.extractor = extractor;
-            this.config = config;
+            this.config = config.History;
```
ozangoktan, line 116: Can this be a property instead?

Extractor/History/HistoryScheduler.cs
```diff
@@ -505,14 +520,27 @@ private void HistoryDataHandler(HistoryReadNode node)
...
+                if (first > dp.SourceTimestamp)
+                {
+                    first = dp.SourceTimestamp;
+                }
+
+                if (last < dp.SourceTimestamp)
+                {
+                    last = dp.SourceTimestamp;
+                }
+                goodDpCount++;
```
ozangoktan, line 538: Is first-last point determination based on only good points? The change block above does filter out unwanted points depending on the configuration, do we want to re-scope the points here?

Extractor/Streamer.cs
```diff
@@ -335,7 +350,10 @@ public void HandleStreamedDatapoint(DataValue datapoint, VariableExtractionState
             }
 
             var buffDps = ToDataPoint(datapoint, node);
-            node.UpdateFromStream(buffDps);
+            if (StatusCode.IsGood(datapoint.StatusCode))
+            {
+                node.UpdateFromStream(buffDps);
+            }
```
ozangoktan, line 356: Similar situation here, we've already checked above, should we not remove this check?

Extractor/Types/UADataPoint.cs
```diff
@@ -88,10 +126,11 @@ public UADataPoint(UADataPoint other, double replacement)
         /// <returns>Array of bytes</returns>
         public byte[] ToStorableBytes()
         {
-            var bytes = new List<byte>(16);
+            var bytes = new List<byte>(20);
             bytes.AddRange(CogniteUtils.StringToStorable(Id));
             bytes.AddRange(BitConverter.GetBytes(Timestamp.ToBinary()));
             bytes.AddRange(BitConverter.GetBytes(IsString));
+            bytes.AddRange(BitConverter.GetBytes(Status.Code));
```
ozangoktan, line 133: Would be a good idea to mention in the changelog that this version of the extractor will discard locally stored datapoints that haven't yet been ingested to CDF.

Extractor/History/HistoryScheduler.cs
einarmo: I think it has to be. Some servers will return a bunch of bad datapoints if they are in a bad state somehow. We can't really trust those when it comes to state.

The exception is if we rely on timestamps for pagination, in which case we have to trust it, but since that's a workaround anyway it's fine.

Extractor/Streamer.cs
einarmo: Same here, we only want to keep state for good points.

ozangoktan (approved):
LGTM 🎸

PR #596, opened 2024-02-06
Improve subscription recreation logging

einarmo:
Also fix an errant callback which should have been removed in 2.25


ozangoktan (approved):
LGTM

PR #605, opened 2024-02-15
Enforce start end time

einarmo:
The extractor works but behaves very weirdly if this is not true, so it's best to enforce it.

This is not really a breaking change, any deployments with this configuration are fundamentally broken.


rsjr (approved):
LGTM (aprob)

PR #610, opened 2024-02-21
Improve history restart reliability a bit

einarmo:
I couldn't reproduce the actual issue here, but I noticed potential for some racyness and the logging was a bit thin. I also don't trust AutoResetEvent, so I'm swapping it for ManualResetEvent.

### DIFF
--- a/Extractor/History/HistoryReader.cs
+++ b/Extractor/History/HistoryReader.cs
@@ -289,7 +289,7 @@ namespace Cognite.OpcUa.History
 
                 if (stateOk != state.IsGood)
                 {
-                    log.LogDebug("History state changed, history will {Event}", state.IsGood ? "Restart" : "Stop");
+                    log.LogDebug("History state changed to good, history will restart");
                     stateChangedEvent.Set();
                 }
             }
@@ -308,7 +308,7 @@ namespace Cognite.OpcUa.History
 
                 if (stateOk != state.IsGood)
                 {
-                    log.LogDebug("History state changed, history will {Event}", state.IsGood ? "Restart" : "Stop");
+                    log.LogDebug("History state changed to bad, history will stop");
                     stateChangedEvent.Set();
                 }
             }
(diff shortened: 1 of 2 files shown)


Extractor/History/HistoryReader.cs
```diff
@@ -276,31 +278,39 @@ private void RestartHistoryInStates()
...
-            {
-                stateChangedEvent.Set();
+                if (state.RemoveIssue(issue))
+                {
+                    log.LogDebug("Removed history issue: {Issue}. New state: {State}", issue, state);
+                }
+
+                if (stateOk != state.IsGood)
+                {
+                    log.LogDebug("History state changed, history will {Event}", state.IsGood ? "Restart" : "Stop");
```
ozangoktan, line 292: Maybe we can just hardcode Restart here, since it should never say Stop.

Extractor/History/HistoryReader.cs
```diff
@@ -276,31 +278,39 @@ private void RestartHistoryInStates()
...
-            {
-                stateChangedEvent.Set();
+                if (state.AddIssue(issue))
+                {
+                    log.LogDebug("Added history issue: {Issue}. New state: {State}", issue, state);
+                }
+
+                if (stateOk != state.IsGood)
+                {
+                    log.LogDebug("History state changed, history will {Event}", state.IsGood ? "Restart" : "Stop");
```
ozangoktan, line 311: Same as above, but the states reversed.

ozangoktan (commented):
Generally looks good, had a minor comment for logging

Extractor/History/HistoryReader.cs
einarmo: It can, the condition is not a check of `state.IsGood`, it's a check of whether `state.IsGood` has changed, it goes both ways.

ozangoktan: The method calls RemoveIssue, and that can only change IsGood from false to true. This block is dependent on there being a change, therefore this can only say Restart.

einarmo: Ah. You're right.

ozangoktan (approved):
LGTM

PR #626, opened 2024-04-22
Include filter

einarmo:
This seems to be the more common use of ignore filters, to explicitly only include a specific set of nodes. This implements that explicitly.


ozangoktan (approved):
LGTM

PR #644, opened 2024-05-08
Null support

einarmo:


mathialo (commented):
Looks good

We need to add the status code thing to the example configs that's shipped with the artifact though.

mathialo (approved):

PR #683, opened 2024-08-20
Log a warning if no variables were found

einarmo:


Extractor/SessionManager.cs
```diff
@@ -457,10 +457,10 @@ private async Task<SessionRedundancyResult> CreateSessionWithRedundancy(IEnumera
                     {
                         if (current != null)
                         {
-                            await current.Session.CloseAsync();
                             current.Session.KeepAlive -= ClientKeepAlive;
                             current.Session.PublishError -= OnPublishError;
                             current.Session.Dispose();
+                            await current.Session.CloseAsync();
```
ozangoktan, line 463: Disposing a session before closing it? That does not seem right.

Extractor/UAExtractor.cs
```diff
@@ -959,6 +959,14 @@ private async Task PushNodes(PusherInput input)
 
             bool initial = input.Variables.Count() + input.Objects.Count() >= State.NumActiveNodes;
 
+            if (initial && input.Variables.Any() && Config.Subscriptions.DataPoints)
```
ozangoktan, line 962: The only way for this if to evaluate true is for there to be any input variables, but the message is saying there are no variables. Is this check correct?

einarmo: Oops, yeah, it should be the inverse.

PR #685, opened 2024-08-21
Don't run docker as root

mathialo:


Dockerfile
```diff
@@ -12,4 +16,10 @@ ENV OPCUA_CERTIFICATE_DIR="/certificates"
 ENV OPCUA_OWN_CERTIFICATE_DIR="/certificates/pki/own"
 ENV OPCUA_CERTIFICATE_SUBJECT="CN=Opcua-extractor, C=NO, S=Oslo, O=Cognite, DC=docker.opcua"
 
+RUN mkdir -p /logs
+RUN chmod -R a+rw /logs
```
einarmo, line 20: You probably have to chmod `/certificates` as well.

mathialo:
I assume that `/logs` is the only directory the running user needs _write_ access to, it that fair?

Dockerfile
mathialo: I checked, and it should have read access, do we need write access as well?

einarmo: Yup

einarmo (approved):

PR #693, opened 2024-08-30
Bump packages and version

einarmo:


manifest.yml
```diff
@@ -67,6 +67,11 @@ schema:
    - "https://raw.githubusercontent.com/"
 
 versions:
+  "2.30.3":
+    description: Fix issue causing error when updating parent IDs.
+    changelog:
+      fixed:
```
ozangoktan, line 73: Would be good to briefly mention fix from https://github.com/cognitedata/dotnet-extractor-utils/pull/458 also

PR #702, opened 2024-09-11
Idm support

einarmo:


ozangoktan (approved):
Commented on the error messages which could be improved, other than that looks good. The SourceInformation class is a nice addition.

PR #704, opened 2024-09-12
Configure and run dotnet format on everything

einarmo:


ozangoktan (commented):
I've added comments on the things it does that I think we are better off without. I would say we should also integrate this to be checked within the build pipeline after disabling some of the rules.

ozangoktan (approved):

PR #739, opened 2024-11-27
More connect improvements

einarmo:
This logic has gotten really hairy. A proper rethink is in order, but this should hopefully tighten up a few holes. We sometimes have leaky and bad behavior when the server is being really unstable, i.e. we connect and it immediately stops working.

This may help a bit.


einarmo:
Closing this in favor of a much larger change later...

PR #747, opened 2024-12-05
Refactor SessionManager

einarmo:
The current solution is just too fragile and unstable. I wanted to avoid doing this, but I didn't see a way forward with the current implementation. This fundamentally changes a few things:

 - Abstract over reconnect/connect methods by creating an `IConnectionSource` interface.
 - Run a single loop to actually invoke `Connect`. This loop is notified from other threads through a condition variable and holds a lock over almost its entire body, which should help simplify the possible traces of the session manager, and make future bugs easier to fix and less likely to occur in the first place.

I have tested this as much as I am able, and of course there are some automated tests that still pass, but the issues we've had with the old implementation are really difficult to reproduce, since they relied heavily on highly specific timing.


Extractor/Connect/SessionManager.cs
```diff
@@ -0,0 +1,556 @@
...
+        {
+            var res = await session.ReadAsync(null, 0, TimestampsToReturn.Neither, new ReadValueIdCollection
+            {
+                new ReadValueId
+                {
+                    NodeId = VariableIds.Server_ServiceLevel,
+                    AttributeId = Attributes.Value
+                }
+            }, token);
+            var dv = res.Results[0];
```
ozangoktan, line 282: Is this guaranteed to not null-reference/index out of range?

Extractor/Connect/SessionManager.cs
```diff
@@ -0,0 +1,556 @@
...
+        }
+
+        private async Task<bool> UpdateServiceLevel(byte newLevel, bool fromConnection)
+        {
+            var oldServiceLevel = CurrentServiceLevel;
+
+            if (newLevel != oldServiceLevel)
+            {
+                log.LogDebug("Server ServiceLevel updated {From} -> {To}", oldServiceLevel, newLevel);
+            }
```
ozangoktan, line 312: else return false? All the rest of the method seems to be dealing with level change.

Extractor/Connect/SessionManager.cs
```diff
@@ -0,0 +1,556 @@
...
+
+            // Fetch the current service level as well, and trigger an update.
+            if (config.IsRedundancyEnabled || config.Redundancy.MonitorServiceLevel)
+            {
+                var sl = await ReadServiceLevel(result.Connection.Session, token);
+                var slResult = UpdateServiceLevel(sl, true);
+            }
+
+            if (connection?.Session != result.Connection.Session)
```
ozangoktan, line 390: Would it make sense to compare by the SessionId property?

Extractor/Connect/SessionManager.cs
```diff
@@ -0,0 +1,556 @@
...
+            if (sub != null)
+            {
+                log.LogError("Unexpected error on publish: {Code}, subscription: {Name}", symId, sub.DisplayName);
+            }
+        }
+
+        private void ClientKeepAlive(ISession sender, KeepAliveEventArgs eventArgs)
+        {
+            client.LogDump("Keep Alive", eventArgs);
+            if (eventArgs.Status == null || !ServiceResult.IsNotGood(eventArgs.Status)) return;
```
ozangoktan, line 469:
```suggestion
            if (eventArgs.Status == null || ServiceResult.IsGood(eventArgs.Status)) return;
```

ozangoktan (commented):
The logic generally looks better, and the interface makes dealing with the connection types much nicer. Had some comments.

Extractor/Connect/DirectConnectionSource.cs
```diff
@@ -0,0 +1,122 @@
...
+                else throw;
+            }
+
+            return false;
+        }
+
+
+        public async Task<ConnectResult> Connect(
+            Connection? oldConnection,
+            bool isConnected,
```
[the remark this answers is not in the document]
einarmo: So, maybe, the problem is that the knowledge of whether we are connected or not is highly contextual. It's one of the major pain points of the OPC-UA SDK. I'll think about it, but it's really hard to do in a way that doesn't cause other problems down the line.

Extractor/Connect/SessionManager.cs
einarmo: By the OPC-UA standard it is, but not by the SDK I think... We should throw if that happens, but I can throw something explicit.

Extractor/Connect/SessionManager.cs
einarmo: That's fair.

Extractor/Connect/SessionManager.cs
einarmo: Not for this code. This is important for guarding against memory leaks, here I actually do care about whether we have a new _Session_ object, not whether we have a new OPC-UA session.

Extractor/Connect/SessionManager.cs
einarmo: Not the same thing! An uncertain keep alive is a failed keep alive.

Extractor/Connect/DirectConnectionSource.cs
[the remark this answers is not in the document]
einarmo: Changing this requires dropping or making the `SessionManagerState` more complex. I really hate that, the state is the thing that makes the session manager logic possible to follow, if that becomes tied to the `Connection` object it makes everything more complex.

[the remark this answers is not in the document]
ozangoktan: I think this is OK for now, we can iterate on it if it makes sense later.

ozangoktan (approved):

PR #755, opened 2025-01-02
Avoid leaking datapoints in the switchover buffer

einarmo:
The purpose of this buffer is to avoid data loss on the "switch" from historical to live data. Recent changes have kept us "idle" in the IsFrontfilling state in a way that we never saw earlier. This change should help avoid leaking memory there in that case.

This whole feature can be removed once we make the state store required...


einarmo:
Actually, should we nuke the whole feature now? If people have issues with loss of data during complex history reads, they really, really should be using the state store anyway.

sagar-thalwar (approved):

toondaey (approved):

PR #761, opened 2025-01-16
Use HistoryReader state for push status as well

einarmo:
We want to restart history if we fail to push data to CDF. We already do this, but now that there's a unified system for managing these restarts, it is better to use that.

### DIFF
--- a/Extractor/History/HistoryReader.cs
+++ b/Extractor/History/HistoryReader.cs
@@ -269,7 +269,9 @@ namespace Cognite.OpcUa.History
             ServerConnection,
             DataPointSubscription,
             EventSubscription,
-            NodeHierarchyRead
+            NodeHierarchyRead,
+            DataPushFailing,
+            EventsPushFailing,
         }
 
         private void RestartHistoryInStates()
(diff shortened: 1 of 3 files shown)


Extractor/FailureBuffer.cs
```diff
@@ -313,7 +313,7 @@ private async Task WriteEventsInflux(IEnumerable<UAEvent> events, CancellationTo
                     if (emitterState == null) continue;
                     eventBufferStates[group.Id] = bufferState = new InfluxBufferState(emitterState);
                 }
-                if (extractor.AllowUpdateState) bufferState.UpdateDestinationRange(group.Range.Min, group.Range.Max);
+                bufferState.UpdateDestinationRange(group.Range.Min, group.Range.Max);
```
einarmo, line 316: Turns out I was a bit overeager to use the `AllowUpdateState` field. This particular state should be updated no matter what...

I really want to nuke this feature too.

mathialo: maybe add a small note about why we're not checking for it here, so it doesn't look like an omission for someone reading it later?

mathialo (approved):

PR #765, opened 2025-01-20
Remove read extracted ranges

einarmo:
Removes the read-extracted-ranges feature completely.

There may be a little room for more refactoring in the UAExtractor class, but not a lot and that can wait for later. This makes state-storage required, sets the state-storage interval by default (you still need to set a location and type), and removes the `read-extracted-ranges` option everywhere.


Test/Integration/EventTests.cs
```diff
@@ -445,88 +444,6 @@ await TestUtils.WaitForCondition(() =>
                 await extractor.DisposeAsync();
             }
         }
-        [Theory]
-        [InlineData(true)]
-        [InlineData(false)]
-        public async Task TestPusherStateRestart(bool backfill)
```
ozangoktan, line 451: This doesn't seem to be a test specific to the state store workaround, is there a reason to remove it?

Extractor/Config/StateStorageConfig.cs
```diff
@@ -55,5 +56,9 @@ public string? Interval
         /// Name of the raw table or litedb store for storing known reference-type nodes, used for detecting deleted nodes.
         /// </summary>
         public string KnownReferencesStore { get; set; } = "known_references";
+
+
+        [YamlIgnore]
+        public bool IsEnabled => Location != null && Database != StorageType.None;
```
ozangoktan, line 62: Location Null or empty? Interval is not 0? I think those also apply.

Test/Integration/DataPointTests.cs
```diff
@@ -686,92 +684,10 @@ await TestUtils.WaitForCondition(() => extractor.State.NodeStates.All(node =>
                 await extractor.DisposeAsync();
             }
         }
-        [Theory]
-        [InlineData(true)]
-        [InlineData(false)]
-        public async Task TestPusherStateRestart(bool backfill)
```
ozangoktan, line 692: Same as the comment on the other test

Extractor/UAExtractor.cs
```diff
@@ -1014,7 +988,7 @@ private async Task PushNodes(PusherInput input)
                 }
             }
 
-            if (StateStorage != null && Config.StateStorage.IntervalValue.Value != Timeout.InfiniteTimeSpan)
+            if (Config.StateStorage.IsEnabled && StateStorage != null)
```
ozangoktan, line 991: Is the interval value no longer applicable? The doc still says 0 value disables it.

Test/Integration/DataPointTests.cs
einarmo: Both of these tests were tests for seeing how the state changed _without_ a state store when the extractor restarted, i.e. how the state was restored from CDF.

Extractor/Config/StateStorageConfig.cs
einarmo: I'll make it null or empty to be safe, but for interval as mentioned above I think expected behavior is to just never write to it.

einarmo: Actually, I'll change this so that setting `StorageType` to not-none and not having a location is a hard error on startup.

schema/state_storage_config.schema.json
```diff
@@ -13,7 +13,7 @@
         "interval": {
             "type": "string",
             "description": "Interval between each write to the buffer file, in seconds. 0 or less disables the state store. Format is as given in [Timestamps and intervals](#timestamps-and-intervals).",
```
ozangoktan, line 15: If 0s setting makes the state store read-only, then this line should be clarified to mean that. At the moment it sounds like it does not read either.

ozangoktan (approved):

PR #766, opened 2025-01-22
Remove entire MQTT pushing mechanism

einarmo:
This was never properly kept up to date with all the new features we added to the extractor, it was never used the way it was intended (only used once or twice in ways it wasn't intended to, with great difficulty), and it was a big maintenance burden.

If we want a feature like this we should design it from the ground up in a completely different way, probably leaving the ingestion up to the user, not writing some crazy bridge application.


toondaey (approved):
🫥

PR #767, opened 2025-01-22
Make update non-optional

einarmo:
When we added support for "updating" metadata in CDF, it was done in a very flexible way.

Now that we write to DM more, it really makes a lot less sense, since depending on where you write there is no way to _not_ update. We had some absolutely insane code to enable "updates" to Raw that can also be removed (no other extractors do anything like that).

In general, making updates non-optional greatly simplifies the extractor, and is almost certainly what pretty much every user wants.

Also identifies and fixes a bug related to the `Source` of arrays, this was never caught due to a hole in a test.

### DIFF
--- a/Extractor/Pushers/PusherUtils.cs
+++ b/Extractor/Pushers/PusherUtils.cs
@@ -80,9 +80,10 @@ namespace Cognite.OpcUa.Pushers
         /// <summary>
         /// Create timeseries update from existing timeseries and new OPC-UA variable.
         /// </summary>
+        /// <param name="config">Extractor configuration object.</param>
+        /// <param name="client">Interface for generating unique IDs using the client context.</param>
         /// <param name="old">Existing timeseries</param>
         /// <param name="newTs">New OPC-UA variable</param>
-        /// <param name="update">Configuration for which fields to update</param>
         /// <param name="nodeToAssetIds">Map from NodeIds to assetIds, necessary for setting parents</param>
         /// <returns>Update object, or null if updating was unnecessary</returns>
         public static TimeSeriesUpdate? GetTSUpdate(
@@ -132,10 +133,10 @@ namespace Cognite.OpcUa.Pushers
         /// <summary>
         /// Create asset update from existing asset and new OPC-UA node.
         /// </summary>
-        /// <param name="extractor">Active extractor, used for building metadata</param>
+        /// <param name="config">Extractor config.</param>
         /// <param name="old">Existing asset</param>
         /// <param name="newAsset">New OPC-UA node</param>
-        /// <param name="update">Configuration for which fields to update</param>
+        /// <param name="extractor">Active extractor, used for building metadata</param>
         /// <returns>Update object, or null if updating was unnecessary</returns>
         public static AssetUpdate? GetAssetUpdate(
             FullConfig config,
--- a/Extractor/Pushers/PusherUtils.cs
+++ b/Extractor/Pushers/PusherUtils.cs
@@ -60,32 +60,6 @@ namespace Cognite.OpcUa.Pushers
             }
         }
 
-        public static JsonElement? CreateRawUpdate(
-            ILogger log,
-            StringConverter converter,
-            BaseUANode node,
-            RawRow<Dictionary<string, JsonElement>>? raw,
-            ConverterType type)
-        {
-            if (node == null) return null;
-            var newObj = node.ToJson(log, converter, type);
-
-            if (newObj == null || newObj.RootElement.ValueKind != JsonValueKind.Object) return null;
-            if (raw == null) return newObj.RootElement;
-
-            var fields = new HashSet<string>();
-            foreach (var row in newObj.RootElement.EnumerateObject())
-            {
-                if (!raw.Columns.TryGetValue(row.Name, out JsonElement value) || value.ToString() != row.Value.ToString())
-                {
-                    return newObj.RootElement;
-                }
-                fields.Add(row.Name);
-            }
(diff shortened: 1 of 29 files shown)


Extractor/Pushers/Writers/TimeseriesWriter.cs
```diff
@@ -71,15 +71,15 @@ protected override IEnumerable<TimeSeriesCreate> BuildTimeseries(
         /// <param name="token">Cancellation token</param>
         /// <returns>Operation result</returns>
         protected override async Task UpdateTimeseries(UAExtractor extractor, IDictionary<string, UAVariable> tsMap,
-                IEnumerable<TimeSeries> timeseries, IDictionary<NodeId, long> nodeToAssetIds, TypeUpdateConfig update, Result result, CancellationToken token)
+                IEnumerable<TimeSeries> timeseries, IDictionary<NodeId, long> nodeToAssetIds, Result result, CancellationToken token)
```
ozangoktan, line 74: param documentation is still there

Extractor/Pushers/PusherUtils.cs
```diff
@@ -116,46 +90,40 @@ private static bool ShouldSetNewMetadata(FullConfig config, Dictionary<string, s
             IUAClientAccess client,
             TimeSeries old,
             UAVariable newTs,
-            TypeUpdateConfig update,
```
ozangoktan, line 119: function doc needs to be updated

Extractor/Pushers/Writers/RawWriter.cs
```diff
@@ -168,90 +153,14 @@ public async Task MarkDeleted(DeletedNodes deletes, CancellationToken token)
         /// <param name="token">Cancellation token</param>
         /// <returns>Operation result</returns>
         private async Task<Result> PushRows<T>(UAExtractor extractor, string database, string table,
-                IReadOnlyDictionary<string, T> rows, ConverterType converter, bool shouldUpdate, CancellationToken token) where T : BaseUANode
```
ozangoktan, line 171: function doc needs to be updated

Extractor/Pushers/Writers/CleanWriter.cs
```diff
@@ -66,7 +66,6 @@ public async Task<bool> PushAssets(
             UAExtractor extractor,
             IDictionary<string, BaseUANode> nodes,
             IDictionary<NodeId, long> nodeToAssetIds,
-            TypeUpdateConfig update,
```
ozangoktan, line 69: same here

Extractor/Pushers/Writers/CleanWriter.cs
```diff
@@ -195,15 +190,15 @@ private async Task<IEnumerable<Asset>> CreateAssets(UAExtractor extractor,
         /// <param name="token">Cancellation token</param>
         /// <returns>Future list of assets</returns>
         private async Task UpdateAssets(UAExtractor extractor, IDictionary<string, BaseUANode> assetMap,
-                IEnumerable<Asset> assets, TypeUpdateConfig update, Result result, CancellationToken token)
```
ozangoktan, line 198: same here

Extractor/Pushers/Writers/BaseTimeseriesWriter.cs
```diff
@@ -44,7 +44,6 @@ public async Task<bool> PushVariables(
             IDictionary<string, UAVariable> timeseriesMap,
             IDictionary<NodeId, long> nodeToAssetIds,
             HashSet<string> mismatchedTimeseries,
-            TypeUpdateConfig update,
```
ozangoktan, line 47: same here

Extractor/Pushers/Writers/RawWriter.cs
```diff
@@ -265,109 +174,16 @@ await UpsertRows(
         /// <param name="token">Cancellation token</param>
         /// <returns>Task</returns>
         private async Task CreateRows<T>(UAExtractor extractor, string database, string table,
-                IReadOnlyDictionary<string, T> dataMap, ConverterType converter, Result result, CancellationToken token) where T : BaseUANode
```
ozangoktan, line 268: same here

Extractor/Pushers/Writers/RawWriter.cs
```diff
@@ -265,109 +174,16 @@ await UpsertRows(
         /// <param name="token">Cancellation token</param>
         /// <returns>Task</returns>
         private async Task CreateRows<T>(UAExtractor extractor, string database, string table,
```
ozangoktan, line 176: Maybe we can call this method 'UpsertRows' to be more in line with what the API is doing.

Test/Integration/NodeExtractionTests.cs
```diff
@@ -1221,6 +1169,7 @@ void Compare(IEnumerable<BaseUANode> nodes, IEnumerable<UAVariable> variables, H
             // Enable types only
             tester.Log.LogInformation("BEGIN TYPE RUN");
             tester.Client.TypeManager.Reset();
+            extractor.State.Clear();
```
ozangoktan, line 1172: This method is only used by tests. Could it obscure bugs down the line, since the extractor itself does not make this call.

einarmo: You could _maybe_ recreate the extractor in some of these tests, except C# doesn't allow reassigning `using` variables, so you'd have to rewrite all the tests that use this approach, I think. They would still share a client.

In practice there is little difference.

ozangoktan (approved):
LGTM

PR #775, opened 2025-01-30
Reduce default concurrency in config.example.yml

amorken:
The default concurrency settings should be conservative and increased when there's a need rather than be high by default and scaled down when they cause problems.

### DIFF
--- a/config/config.example.yml
+++ b/config/config.example.yml
@@ -386,7 +386,7 @@ cognite:
         # Maximum number of parallel requests per timeseries operation
         time-series: 20
         # Maximum number of parallel requests per assets operation
-        assets: 20
+        assets: 1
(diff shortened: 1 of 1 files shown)


einarmo:
This won't actually fix the problem, as they are hard-coded elsewhere. We can look into reducing them, I agree they are way too high, but I disagree with setting them to 1. If a user produces enough events to ever use more than a single parallel write, they probably want to have at least a few parallel writes.

It's not always easy to understand that the extractor isn't keeping up, and users _do not read documentation for extractors_. This is not the same as API docs or most of our other technical documentation, as the level of technical know-how is often way lower with extractors.

I've personally never seen a user correctly change these settings. I've seen them set to 0, and to 1000. If we can set sane defaults that rarely need changing I would much rather do that.

PR #779, opened 2025-02-10
Fix sanitation of DM external IDs

einarmo:
The way we did it was dumb and also didn't work for every case. This is a much better way to do it.

Also remove the index on `TypeHierarchy`. I'm reasonably certain this was a breaking change in DM...

### DIFF
--- a/Extractor/Pushers/FDM/FDMUtils.cs
+++ b/Extractor/Pushers/FDM/FDMUtils.cs
@@ -6,10 +6,16 @@ namespace Cognite.OpcUa.Pushers.FDM
     public static class FDMUtils
     {
         private static readonly Regex extIdRegex = new Regex("^[a-zA-Z]([a-zA-Z0-9_]{0,253}[a-zA-Z0-9])?$", RegexOptions.Compiled);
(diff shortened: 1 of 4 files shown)


Extractor/Pushers/FDM/FDMUtils.cs
```diff
@@ -6,10 +6,16 @@ namespace Cognite.OpcUa.Pushers.FDM
...
 
         public static string SanitizeExternalId(string raw)
         {
-            var clean = raw.Replace('-', '_').Replace(' ', '_').Replace("<", "").Replace(">", "").Replace("/", "_").Replace("+", "_").Replace(".", "_").Replace("@", "_");
+            var clean = illegalSymbol.Replace(raw, match => match.Value switch
+            {
+                "<" => "",
+                ">" => "",
+                _ => "_"
+            }).TrimEnd('_');
```
ozangoktan, line 18: Can we add a simple unit test around SanitizeExternalId?

einarmo: Added a simple unit test. This is also tested indirectly in some other tests.

ozangoktan (approved):
LGTM

PR #782, opened 2025-02-13
Make status codes enabled always

einarmo:
Also removes `NonFiniteReplacement` which no longer does anything.

This should be a relatively clean removal.

### DIFF
--- a/Extractor/Streamer.cs
+++ b/Extractor/Streamer.cs
@@ -357,20 +357,6 @@ namespace Cognite.OpcUa
                     log.LogDebug("Bad streaming datapoint: {BadDatapointExternalId} {SourceTimestamp}. Value: {Value}, Status: {Status}",
                         node.Id, datapoint.SourceTimestamp, datapoint.Value, ExtractorUtils.GetStatusCodeName((uint)datapoint.StatusCode));
                 }
-
-                switch (config.Extraction.StatusCodes.StatusCodesToIngest)
-                {
-                    case StatusCodeMode.All:
-                        break;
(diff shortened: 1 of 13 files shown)


Extractor/Streamer.cs
```diff
@@ -390,10 +376,6 @@ public void HandleStreamedDatapoint(DataValue datapoint, VariableExtractionState
                 node.UpdateFromStream(buffDps);
             }
 
-            if ((extractor.StateStorage == null || config.StateStorage.IntervalValue.Value == Timeout.InfiniteTimeSpan)
-                 && (node.IsFrontfilling && datapoint.SourceTimestamp > node.SourceExtractedRange.Last
-                    || node.IsBackfilling && datapoint.SourceTimestamp < node.SourceExtractedRange.First)) return;
```
ozangoktan, line 395: Is there a reason to remove this check? Seems unrelated to status codes.

ozangoktan (commented):
Example config still refers to nanReplacement https://github.com/cognitedata/opcua-extractor-net/blob/de4950fc66794c4848366a17db445ef895007eac/config/config.example.yml#L225

Extractor/Streamer.cs
einarmo: Should have been removed in the state-store required PR, and I think caused some unexpected behavior when I was testing this.

einarmo: There's some more that should be removed here... I'll clean it up in a bit.

ozangoktan (approved):

PR #783, opened 2025-02-14
Remove history buffer thingy

einarmo:
This was used to bridge the gap between history end and stream start. Now that state store is required when history is enabled, we no longer need this at all.


ozangoktan (approved):
nice!

PR #784, opened 2025-02-14
Always enable extra data to Raw

einarmo:
This isn't actually that much data, and has no real cost on startup. Including this info makes setting up raw-node-buffer easier, and is one less thing to think about when configuring the extractor.


ozangoktan (approved):
LGTM

PR #795, opened 2025-03-18
Register inverse mapping for the server node

einarmo:
This is used when events are stored in the failure buffer, there's a potential issue here that may cause events to not be properly restored.


ozangoktan (approved):

larshagencognite (approved):
risk review ok

PR #799, opened 2025-03-26
Bump utils and SDK

einarmo:
Also use the new TimestampWrapper type from the utils, which should help simplify config.

Docs are updated to match.

### DIFF
--- a/Test/Integration/LauncherTests.cs
+++ b/Test/Integration/LauncherTests.cs
@@ -505,13 +505,6 @@ version: 1
                 method.Invoke(typeof(ExtractorStarter), new object[] { log, config, setup, options, "config", services }));
             Assert.Equal("Invalid config: Do not use events. as id-prefix, as it is used internally", exc.InnerException.Message);
 
-            // Invalid history start time
-            config.Extraction.IdPrefix = null;
-            config.History.StartTime = "2d-agoo";
-            exc = Assert.Throws<TargetInvocationException>(() =>
-                method.Invoke(typeof(ExtractorStarter), new object[] { log, config, setup, options, "config", services }));
-            Assert.Equal("Invalid config: Invalid history start time: 2d-agoo", exc.InnerException.Message);
-
             // Missing opc-ua xml config
             config.History.StartTime = null;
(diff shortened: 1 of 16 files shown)


Test/Integration/LauncherTests.cs
```diff
@@ -505,15 +505,8 @@ public void TestVerifyAndBuildConfig()
...
-            // Invalid history start time
-            config.Extraction.IdPrefix = null;
-            config.History.StartTime = "2d-agoo";
-            exc = Assert.Throws<TargetInvocationException>(() =>
-                method.Invoke(typeof(ExtractorStarter), new object[] { log, config, setup, options, "config", services }));
-            Assert.Equal("Invalid config: Invalid history start time: 2d-agoo", exc.InnerException.Message);
-
             // Missing opc-ua xml config
-            config.History.StartTime = null;
+            config.Extraction.IdPrefix = null;
```
ozangoktan, line 509: What is this line doing?

psalaberria002 (commented):
This PR does more than bumping utils and SDK.

Change it accordingly.

Could this PR be done in two instead? One for bumping deps, and one for making the timestampwrapper change?

einarmo:
> This PR does more than bumping utils and SDK.
> 
> Change it accordingly.
> 
> Could this PR be done in two instead? One for bumping deps, and one for making the timestampwrapper change?

Why? It's not a big change, and the changes are coupled (the timestamp wrapper is in the new version of the SDK). This is not getting deployed anywhere, as I did not bump the version, so is there any reason why I should split it and need to wait another week for reviews?

psalaberria002:
> > This PR does more than bumping utils and SDK.
> > Change it accordingly.
> > Could this PR be done in two instead? One for bumping deps, and one for making the timestampwrapper change?
> 
> Why? It's not a big change, and the changes are coupled (the timestamp wrapper is in the new version of the SDK). This is not getting deployed anywhere, as I did not bump the version, so is there any reason why I should split it and need to wait another week for reviews?

It wasn't obvious from the PR description. Next time write that down in the description to avoid confusion. Thanks.

psalaberria002:
:unicorn:

PR #800, opened 2025-03-31
Fix the opcua_subscriptions metric

einarmo:
It's just very inaccurate. We can improve it a lot by making it labeled per subscription, since we now (we didn't when it was added, years ago), have unique names for each subscription we create.

The removed line from the test is unfortunate, but we use a _horrific_ hack to pull the metric values in tests, and it doesn't work on labeled metrics, and I REALLY do not want to fix that.

### DIFF
--- a/Test/Unit/UAClientTest.cs
+++ b/Test/Unit/UAClientTest.cs
@@ -1066,7 +1066,6 @@ namespace Test.Unit
                 }
                 tester.Config.Source.SubscriptionChunk = 1000;
             }
-            Assert.True(CommonTestUtils.TestMetricValue("opcua_subscriptions", 2000));
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


Test/Unit/UAClientTest.cs
```diff
@@ -1066,7 +1066,6 @@ await TestUtils.WaitForCondition(() => dps.Count == 4000, 5,
                 }
                 tester.Config.Source.SubscriptionChunk = 1000;
             }
-            Assert.True(CommonTestUtils.TestMetricValue("opcua_subscriptions", 2000));
```
ozangoktan, line 1069: Can we replace this test to one that gets the subscription by name instead of removing it, or is there one doing that already?

einarmo: It's just very, very hard. The existing code is nightmarish, and it digs deep into the internals of the prometheus library (which I believe _deliberately_ makes it hard to fetch the values of metrics, the maintainers kinda suck). It's probably possible to actually fetch the metric by label, but it's another horrible journey of ten layers of reflection and I really don't think it's good to put more stuff like that in the extractor.

ozangoktan: 👍🏼

ozangoktan (approved):

simenl (approved):
:unicorn:

PR #804, opened 2025-04-03
Set event ID on records

einarmo:
External ID is now required, make sure to set it on new records.

Like legacy events, just use the event ID.

### DIFF
--- a/Extractor/Extractor.csproj
+++ b/Extractor/Extractor.csproj
@@ -13,15 +13,13 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="AdysTech.InfluxDB.Client.Net.Core" Version="0.25.0" />
-    <PackageReference Include="Cognite.ExtractorUtils" Version="1.31.0" />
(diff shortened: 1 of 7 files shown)


Extractor/Extractor.csproj
```diff
@@ -14,6 +14,7 @@
   <ItemGroup>
     <PackageReference Include="AdysTech.InfluxDB.Client.Net.Core" Version="0.25.0" />
     <PackageReference Include="Cognite.ExtractorUtils" Version="1.31.0" />
+    <PackageReference Include="CogniteSdk" Version="4.15.0" />
```
ozangoktan, line 17: Push https://github.com/cognitedata/dotnet-extractor-utils/pull/515 through and bump extractor-utils instead?

einarmo: Then we need to bump the utils, but fine...

ozangoktan (approved):

psalaberria002 (approved):
:unicorn:

PR #808, opened 2025-04-22
Fix batching in the failure buffer

einarmo:
Also makes it configurable, which is easy to do in this case. I wrote a couple of tests to verify that we actually batch.

### DIFF
--- a/Extractor/FailureBuffer.cs
+++ b/Extractor/FailureBuffer.cs
@@ -216,7 +216,7 @@ namespace Cognite.OpcUa
             if (points == null || !points.Any() || pointRanges == null || !pointRanges.Any() || fullConfig.DryRun) return true;
 
             points = points.GroupBy(pt => pt.Id)
-                .Where(group => !extractor.State.GetNodeState(group.Key)?.FrontfillEnabled ?? false)
+                .Where(group => extractor.State.GetNodeState(group.Key)?.FrontfillEnabled == false)
                 .SelectMany(group => group)
                 .ToList();
 
@@ -431,8 +431,7 @@ namespace Cognite.OpcUa
                 {
                     var points = new List<UADataPoint>();
 
-                    int count = 0;
-                    while (!token.IsCancellationRequested && count < 1_000_000)
+                    while (!token.IsCancellationRequested && points.Count < config.DatapointsBatch)
                     {
                         var dp = UADataPoint.FromStream(stream);
                         if (dp == null)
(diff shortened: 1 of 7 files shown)


toondaey (approved):
👌🏾

manifest.yml
```diff
@@ -67,6 +67,14 @@ schema:
    - "https://raw.githubusercontent.com/"
 
 versions:
+  "2.35.0":
+    description: Fix batching of datapoints and events from the failure buffer, and make batch sizes configurable.
+    changelog:
+      fixed:
+        - Fix batching of datapoints and events from the failure buffer.
+        - Fix bug causing ID and type definition ID based filtering to not work properly.
```
dmivankov, line 75: test case is missing for this

Extractor/FailureBuffer.cs
```diff
@@ -216,7 +216,7 @@ public async Task<bool> WriteDatapoints(IEnumerable<UADataPoint> points, IDictio
             if (points == null || !points.Any() || pointRanges == null || !pointRanges.Any() || fullConfig.DryRun) return true;
 
             points = points.GroupBy(pt => pt.Id)
-                .Where(group => !extractor.State.GetNodeState(group.Key)?.FrontfillEnabled ?? false)
+                .Where(group => extractor.State.GetNodeState(group.Key)?.FrontfillEnabled == false)
```
dmivankov, line 219: What's the change here? Afaik `!null == null` in C#
new condition is only `true` for `false` on the left
old condition is `false` for `null`. `true` for `false` and `false` for `true`, so exactly the same?

manifest.yml
einarmo: Different PR, #809.

Extractor/FailureBuffer.cs
einarmo: Yep. It's just a lot clearer IMO.

Extractor/FailureBuffer.cs
```diff
@@ -431,8 +431,7 @@ private async Task<bool> ReadDatapointsFromFile(IEnumerable<IPusher> pushers, Ca
                 {
                     var points = new List<UADataPoint>();
 
-                    int count = 0;
-                    while (!token.IsCancellationRequested && count < 1_000_000)
+                    while (!token.IsCancellationRequested && points.Count < config.DatapointsBatch)
                     {
                         var dp = UADataPoint.FromStream(stream);
                         if (dp == null)
```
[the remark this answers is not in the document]
einarmo: The condition on the outer `do ... while` is `!final && !token.IsCancellationRequested`, so once we hit the `dp == null` we will not enter a new iteration.

Extractor/FailureBuffer.cs
einarmo: !null is not in fact a thing in C# at all, `??` has higher operator precedence, but it is _not_ obvious.

dmivankov:
risk review ok

PR #821, opened 2025-05-20
No longer require groups:List and projects:List

einarmo:
We don't technically need these, and they have caused confusion in the past, especially now that they are only used if we _lose_ connection to CDF.

For V3 we are likely dropping this entirely in the utils. Since this is purely making the extractor require _fewer_ ACLs, it is a non-breaking change.


ozangoktan (approved):

larshagencognite (approved):
🦄

PR #823, opened 2025-05-22
Add options to adjust MaxNotificationsPerPublish

einarmo:
This is a server option, in practice, so all we do is send this number to the server. Because the size of datapoint and event notifications can be so different, I added a separate number for the two subscriptions.

We've rarely actually seen an issue with this, servers are free to send however many notifications they want.

### DIFF
--- a/Extractor/Config/SubscriptionConfig.cs
+++ b/Extractor/Config/SubscriptionConfig.cs
@@ -108,11 +108,11 @@ namespace Cognite.OpcUa.Config
         public bool RecreateStoppedSubscriptions { get; set; } = true;
 
         /// <summary>
-        /// Maximum number of event notifications per publish.
+        /// Maximum number of datapoint notifications per publish.
         /// This is the maximum number of datapoints that can be sent in a single publish response
         /// from datapoint subscriptions.
         /// </summary>
-        public uint MaxDataNotificationsPerPublish { get; set; } = 0;
+        public uint MaxDatapointNotificationsPerPublish { get; set; } = 0;
 
         /// <summary>
         /// Maximum number of event notifications per publish.
--- a/Extractor/Config/SubscriptionConfig.cs
+++ b/Extractor/Config/SubscriptionConfig.cs
@@ -108,6 +108,20 @@ namespace Cognite.OpcUa.Config
         public bool RecreateStoppedSubscriptions { get; set; } = true;
 
         /// <summary>
+        /// Maximum number of event notifications per publish.
+        /// This is the maximum number of datapoints that can be sent in a single publish response
+        /// from datapoint subscriptions.
+        /// </summary>
+        public uint MaxDataNotificationsPerPublish { get; set; } = 0;
+
+        /// <summary>
+        /// Maximum number of event notifications per publish.
+        /// This is the maximum number of events that can be sent in a single publish response
+        /// from event subscriptions.
+        /// </summary>
+        public uint MaxEventNotificationsPerPublish { get; set; } = 0;
+
+        /// <summary>
         /// Optional grace period for recreating stopped subscriptions.
         /// Defaults to 8 * Publishing interval
         /// </summary>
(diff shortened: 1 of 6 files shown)


Extractor/Config/SubscriptionConfig.cs
```diff
@@ -107,6 +107,20 @@ public class SubscriptionConfig : SubscriptionInstanceConfig
         /// </summary>
         public bool RecreateStoppedSubscriptions { get; set; } = true;
 
+        /// <summary>
+        /// Maximum number of event notifications per publish.
```
ozangoktan, line 111:
```suggestion
        /// Maximum number of datapoint notifications per publish.
```

Extractor/Config/SubscriptionConfig.cs
```diff
@@ -107,6 +107,20 @@ public class SubscriptionConfig : SubscriptionInstanceConfig
...
+        /// This is the maximum number of datapoints that can be sent in a single publish response
+        /// from datapoint subscriptions.
+        /// </summary>
+        public uint MaxDataNotificationsPerPublish { get; set; } = 0;
```
ozangoktan, line 115: Is the name something mandated by OPCUA standard? Otherwise I think we should clarify this as MaxDatapointNotificationsPerPublish. Should also change example config, schema and manifest accordingly.

Extractor/Subscriptions/BaseCreateSubscriptionTask.cs
```diff
@@ -130,13 +130,26 @@ private async Task<Subscription> EnsureSubscriptionExists(ILogger logger, Sessio
...
+                if (SubscriptionName.IsEvents())
+                {
+                    maxNotificationsPerPublish = config.Subscriptions.MaxEventNotificationsPerPublish;
+                }
+                else if (SubscriptionName.IsDataPoints())
+                {
+                    maxNotificationsPerPublish = config.Subscriptions.MaxDataNotificationsPerPublish;
+                }
+                // For general subscriptions, the default value of 0 is going to be fine. In practice,
+                // we will never receive that many notifications for those, since they are very limited in size.
```
ozangoktan, line 143: This comment confused me a bit
1- Sounds like we will never recieve as many as 0 notifications
2- Can we clarify size of what we are referring to
```suggestion
                // For general subscriptions, the default value of 0 is going to be fine. In practice,
                // we will never receive many notifications for datapoints or events, since they are very limited in size.
```

Extractor/Subscriptions/SubscriptionStateCache.cs
```diff
@@ -28,6 +28,16 @@ public static string Name(this SubscriptionName name)
...
+
+        public static bool IsEvents(this SubscriptionName name)
+        {
+            return name == SubscriptionName.Events;
+        }
+
+        public static bool IsDataPoints(this SubscriptionName name)
+        {
+            return name == SubscriptionName.DataPoints;
+        }
```
ozangoktan, line 40: Should these exist? We are just comparing enums. Is there plans to add more logic here?

schema/subscription_config.schema.json
```diff
@@ -89,6 +89,18 @@
             "default": "-1",
             "description": "Grace period for recreating stopped subscriptions. If this is negative, default to 8 * publishing-interval. Format is as given in [Timestamps and intervals](#timestamps-and-intervals)."
         },
+        "max-data-notifications-per-publish": {
+            "type": "integer",
+            "description": "Maximum number of data notifications per publish. Some servers do not properly support this. The default is 0, which means that the server will decide.",
```
ozangoktan, line 94:
```suggestion
            "description": "Maximum number of datapoint notifications per publish. Some servers do not properly support this. The default is 0, which means that the server will decide.",
```

Extractor/Subscriptions/SubscriptionStateCache.cs
einarmo: I just wanted to not hard-code comparisons of `SubscriptionName` in the base class. This way we can change it centrally if we need to.

Extractor/Subscriptions/BaseCreateSubscriptionTask.cs
einarmo: Rewrote the comment

dmivankov:
:unicorn:

PR #824, opened 2025-05-27
Set MaxMonitoredItemsPerCall on the client

einarmo:
We try to do our own batching here, but the client doesn't expose an interface to allow us to do that. Apparently the current way we do it can fail, or create way too many monitored items to be created at a time. This should hopefully protect against this case.

I still don't know exactly how this can happen, it should be impossible, since we always create monitored items when we add them.


ozangoktan (approved):

larshagencognite (approved):
🦄

PR #852, opened 2025-08-20
Add support for writing to timeseries as JSON in CDF

121yaseen:
- Creating a release for adding support for writing to timeseries as JSON in CDF

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -67,6 +67,11 @@ schema:
    - "https://raw.githubusercontent.com/"
 
 versions:
+  "2.37.2":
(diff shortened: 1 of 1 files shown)


manifest.yml
```diff
@@ -67,6 +67,11 @@ schema:
    - "https://raw.githubusercontent.com/"
 
 versions:
+  "2.37.2":
```
einarmo, line 70:
```suggestion
  "2.38.0":
```

This is a feature release, and needs a semi-major bump.

PR #870, opened 2025-09-30
Don't "Un-initialize" pushers that fail during rebrowse

einarmo:
It doesn't really make any sense to go from being initialized to _not_ being initialized, and it causes some really unfortunate issues where datapoints might get backed up.

This would have been problematic a little while ago, but we now gracefully handle nodes that are missing, so creating a subscription before we create the node isn't so terrible, and it's a corner case either way.

This could probably use a better solution in the future, but it gets _really_ complicated if we need to deal with the ingestion status of each node independently.

Also, while creating a test for this, I found that we don't catch errors during `Init`, which we should.


ozangoktan (approved):
LGTM

